PR #6, opened 2022-05-09
2/reuse text components

Cali0707:
Refactored the text components out of the old repo. Some notes:
1. SCSS is pretty sick, especially the way we can use it to extend classes instead of just applying multiple classes to one component
2. I used [ladle](https://ladle.dev/docs/) to handle documenting of components as it is easy to use, fast, and lightweight
3. To keep the code more organized it is now organized into `src/components/{componentGroup}/{SpecificComponent}/SpecificComponent.jsx`, `src/components/{componentGroup}/{SpecificComponent}/SpecificComponent.scss`, and `src/components/{componentGroup}/{SpecificComponent}/SpecificComponent.stories.jsx` -> I am hoping that using this more structured file system along with ladle can help us find components more quickly


jameskokoska (approved):
Everything looks good!

Freeassassin (approved):

PR #24, opened 2022-05-15
Setup backend

Cali0707:
In this PR, I set up most of the backend dependencies, and refactored the frosh registration route from last year. This route was chosen because it showcases nearly all of the new code architecture, minus middleware.


lukewarmtemp (approved):
Looks good

PR #47, opened 2022-05-23
Setup swagger js docs, as well as swagger UI with basic auth

### LINKED ISSUE #11: Setup swagger jsdoc on backend

Cali0707:
Resolves #11 

To test this, ensure you are using the updated `.env` file. Run the app in docker, and go to `http://localhost:5001/docs` in your browser. You should be asked to sign in. Use the `BASIC_AUTH_USER` and `BASIC_AUTH_PASS` values from the updated `.env` file to sign in. Once signed in, you should see the swagger documentation for `/frosh/register`.


server/src/components.yaml
```diff
@@ -0,0 +1,74 @@
+components:
+  schemas:
+    NewFrosh:
+      type: object
+      required:
+        - name
```
jameskokoska, line 6: Wouldn't this be email? I guess we can figure this out after registration gets back to us though

jameskokoska (approved):
Looks good!

Freeassassin (approved):
Looks good. Ran ok on my machine 👍

server/src/components.yaml
Cali0707: I'm currently planning on figuring out what is required after registration gets back to us, for now this is more of a proof of concept that we can have required parameters

PR #49, opened 2022-05-24
added color scheme, default font helvetica

jameskokoska:
Added color scheme:
image

@natapokie I changed the default font to Helvetica, works on most browsers by default without any font files (avoids times new roman default)


natapokie (approved):
Looks good! Thanks James!

Cali0707 (approved):
LGTM 👍

PR #51, opened 2022-05-25
middleware for error response

natapokie:
will send appropriate status code and message to front end, depending on `err.message` property.
more error messages will be added as needed.
currently don't know how to test if it works

### DIFF
--- a/server/src/middlewares/errorResponseMiddleware.js
+++ b/server/src/middlewares/errorResponseMiddleware.js
@@ -1,6 +1,6 @@
 function errorResponseMiddleware (err, req, res, next ) {
 
-    var statusCode, errorMessage;
+    let statusCode, errorMessage;
 
     if (err.message === "INVALID_EMAIL") {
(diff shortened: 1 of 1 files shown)


Freeassassin (dismissed):

jameskokoska (dismissed):

server/src/middlewares/errorResponseMiddleware.js
```diff
@@ -0,0 +1,16 @@
+function errorResponseMiddleware (err, req, res, next ) {
+
+    var statusCode, errorMessage;
```
Cali0707, line 3:
```suggestion
    let statusCode, errorMessage;
```

Cali0707 (changes requested):
This looks great - I requested one small change. Basically `var` variables are globally scoped so you can run into strange behaviour when using them. For this reason, we generally prefer using `let` to define variables that have to change

Cali0707 (approved):

PR #52, opened 2022-05-25
Set up storybook

Cali0707:
Added storybook to the frontend and updated the existing stories files to work.

To test, navigate to the `client` folder and run `yarn storybook`. You should be able to see two components in the storybook: a Header component and a Text component

@jameskokoska I have not yet added storybook to Docker, as it seems fairly isolated and separate. Lmk if you want it added


jameskokoska (approved):
Perfect thanks!

PR #68, opened 2022-05-27
12 find new email api

### LINKED ISSUE #12: Find new Email API
The scope of this issue is to:
1. Research alternative email APIs and compare them
2. Recommend 1-2 options
3. After consulting on which API to use, implement some functions which allow us to send emails with the new API. These functions should allow for one argument to determine which template is going to be used, and another argument to specify the values of any variables in the template

Freeassassin:
Resolves #12.

The following functions were added: 

`sendSimpleEmail(emailContent, toAddresses)`

A function which can receive a JS object containing either text or html and a list of email addresses and comprise and send the email(s).

`createTemplate(templateName, html, subject, text)`

A function to create and save an email template with fillable variables in the AWS dashboard.

`sendTemplateEmail(templateData, templateName, toAddresses)`

A function to send a templated email with filled in variables.

`sendBulkTemplateEmail(bulkEmailEntries, templateName, templateData)`

A function to send templated emails with filled in variables to a wide range of addresses with different variables for each person.

`sendRawEmail(MIMEstring, toAddresses)`

A function to send a raw MIME formatted messages, which allows for the attachment of files onto the email. Currently not working properly.


Emily9023 (dismissed):

Cali0707 (changes requested):
TODO:
- [ ] Refactor all of the functions to format the data properly, instead of requiring others to do this based off of the AWS docs
- [ ] Fix the jsdoc comments to use `array` instead of `list` and to specify return types
- [ ] Set it up to handle variable from addresses
- [x] Switch to SDK v3 instead of SDK v2

PR #70, opened 2022-05-30
39 navbar component draft and questions

natapokie:
currently there are two navbar components, one for desktop and one for mobile, it's almost done, except for a few issues!
the storyboard also works, but the style looks a little different... 😅

**Questions:**
1. how to make underline stay when you click a navbar link?
    `:active  {} and :vistied{}` do not seem to work.... 
3. for frosh profile, when a frosh is logged in, how to connect initials to the profile icon (top right)? and how to make "Frosh Profile" button appear, and "Login" and "Register" buttons disappear?
4. how to check for mobile and desktop view, and render different components?
    tried using `react-device-detect`, with `<BrowserView></BrowserView>` and `<MobileView></MobileView>` but most likely implemented it wrong
6. how to import svg icons from fontawesome?

**To Do:**
- [ ] make the frosh profile (top right) unclickable (?), note: currently links to `/profile`
- [ ] make f!rosh logo (placeholder in top left) clickable and links to home(?)

**To Test:**
`<Navbar></Navbar>` for desktop viewing (comment or uncomment line in App.jsx), should look like below, 
image
- hover over home, about, and faq to make a purple underline appear
- hovering over buttons and profile makes buttons smaller
- clicking links should take you to the corresponding page

`<NavbarMobile></NavbarMobile>` for mobile viewing should look like below, 
image
- squares are placeholders for icons


jameskokoska:
> how to make underline stay when you click a navbar link?
:active {} and :vistied{} do not seem to work....

Try using a prop (selectedPage) 
If the selected page is "Home", for example, highlight that button
Aside:
Active will hold those styles while the link is actively clicked (try a click and hold it down to see those styles)
Visited style will hold those styles if that link has been previously visited (for example Google shows a Purple link instead of blue)

> for frosh profile, when a frosh is logged in, how to connect initials to the profile icon (top right)? and how to make "Frosh Profile" button appear, and "Login" and "Register" buttons disappear?

Don't worry about this, but you would use a prop
Pass in a bool prop (isLoggedIn) show different buttons depending on its value
You can use an inline if statement to show different things based on the prop

```js
isLoggedIn ? <LogInButton/> : <div/>
```

Do the same for the initials, pass in a String prop with the initials

> how to check for mobile and desktop view, and render different components?
tried using react-device-detect, with <BrowserView></BrowserView> and <MobileView></MobileView> but most likely implemented it wrong

Use CSS, it's probably the easiest way
Wrap each component in a parent div
```html
<>
  <div className="navbar-desktop">
    <Navbar/>
  </div>
  <div className="navbar-mobile">
    <NavbarMobile/>
  </div>
</>
```
```css
.navbar-desktop{
  display:unset;
  @include devices(tablet) {
    display:none;
  }
}
.navbar-profile{
  display:none;
  @include devices(tablet) {
    display:unset;
  }
}
```



> how to import svg icons from fontawesome?

Something like: src={require("../../../assets/icons/exit.svg").default}

Cali0707 (changes requested):
Can you refactor these into one component that uses a breakpoint to determine how to render itself? Otherwise we will need to implement this logic on every page we create. Feel free to reach out to me or the team for assistance with this, as it looks like it may be a larger refactor

jameskokoska:
> Can you refactor these into one component that uses a breakpoint to determine how to render itself? Otherwise we will need to implement this logic on every page we create. Feel free to reach out to me or the team for assistance with this, as it looks like it may be a larger refactor

Yep @natapokie you can try something like I suggested above to keep it as one component

```html
<>
  <div className="navbar-desktop">
    <Navbar/>
  </div>
  <div className="navbar-mobile">
    <NavbarMobile/>
  </div>
</>
```
```css
.navbar-desktop{
  display:unset;
  @include devices(tablet) {
    display:none;
  }
}
.navbar-profile{
  display:none;
  @include devices(tablet) {
    display:unset;
  }
}
```

natapokie:
Thank you @jameskokoska and @Cali0707 !!

package.json
```diff
@@ -16,6 +16,12 @@
     "lint-staged": "^12.4.1"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-free": "^6.1.1",
```
jameskokoska, line 19: If these packages aren't being used anymore, they should be removed

jameskokoska (changes requested):

client/src/scssStyles/_mixins.scss
```diff
@@ -5,7 +5,7 @@
     }
   }
   @if $breakpoint == mobile {
-    @media only screen and (max-width: 599px) {
+    @media only screen and (max-width: 767px) {
```
natapokie, line 8: there was some weird formatting (it wasn't following the css for desktop or mobile) when the width was between 767-600px , so I just changed this for now...! @jameskokoska or @Cali0707 you might want to take a look at this?

client/src/App.css
```diff
@@ -16,6 +16,9 @@
   --text-dark: #000000;
   --text-light: #f3f3f3;
   --text-muted: #c2c2c2;
+
+  --navbar-bg: #ebebeb;
+  --navbar-underline: #e2e2e2;
```
natapokie, line 21: also added a few more colors for the navbar!

client/src/scssStyles/_mixins.scss
jameskokoska: Yep no problem
I think it's because you would need to implement tablet styles before mobile would kick in
@Cali0707 I was thinking we just set them to the same, only design for mobile and desktop should be okay for now
(we can just keep them at the same value)

package.json
natapokie: for some reason, i don't see the packages on my local version anymore, and it says that the branch is up to date...

jameskokoska: Did you make sure you added the files to tracking? go into root directory and `git add *` and commit

natapokie: i am so sorry, i was looking at the .json file in the client directory, i will remove these right away!

jameskokoska (dismissed):

client/src/App.jsx
```diff
@@ -21,7 +27,12 @@ const TransitionRoutes = () => {
   const location = useLocation();
   return (
     <TransitionGroup>
-      <Navbar />
+      <Navbar></Navbar>
+      {/* <Button></Button> */}
+      {/* <NavbarDesktop></NavbarDesktop> */}
+      {/* <NavbarStorybook></NavbarStorybook> */}
+      {/* <TestComponent></TestComponent> */}
+
```
Cali0707, line 35:
```suggestion

```

Cali0707 (approved):
Looks great! Thanks for refactoring this into a new component

PR #89, opened 2022-06-01
added timeline

jameskokoska:
image

### DIFF
--- a/client/src/components/timeline/Timeline/Timeline.jsx
+++ b/client/src/components/timeline/Timeline/Timeline.jsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './Timeline.scss';
+
+const Timeline = ({ dates, onClick }) => {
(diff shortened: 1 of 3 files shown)


natapokie (approved):
looks good!! the scrolling in mobile is very cool !!
just one small thing, the end of the yellow line (far right) in mobile is different than the right side, it pops out a little, once again i'm not sure if this is storybook formating...?
image

and i also tried running the front end with the timeline component, but I got a few errors, it might be an issue on end but I just wanted to let you know!
image

jameskokoska:
Also you'll need to pass in the correct list of objects into the props otherwise it may show an attribute as undefined
An example can be found in the storybook

natapokie:
> Also you'll need to pass in the correct list of objects into the props otherwise it may show an attribute as undefined An example can be found in the storybook

ohhh i see i see!

Cali0707 (approved):
LGTM 👍

Issue #31, opened 2022-05-17
Accordion container component


jameskokoska:
From last year:

Accordion collapsable container
header: string
children: components
position: string (optional) ('top', 'middle', 'bottom')
initialOpen: boolean (default:false) (optional)
accessible functions via ref: ref.setAccordionState(boolean)

![image](https://user-images.githubusercontent.com/50821962/168833552-adfdc4c1-fed7-4dd0-bfa3-6448a13cd34a.png)

PR #91, opened 2022-06-01
created accordion component with animation

lukewarmtemp:
#31 

Couldn't figure out initialOpen with the current way I'm doing the animations, but I can revisit it's something important that we really need. Accordion opens both with onClick() and whenever the toggle prop value changes. Note: open/closed state is independent of what the value is set to. Can make an adjustment to open/close corresponding to toggle=true/false but some changes also need to be made in onClick() function to ensure that we don't have a case where the accordion is closed but toggle is set to true and vice versa.


lukewarmtemp:
Also, I'll get the images of what it looks/prove that it's working sometime tomorrow

client/package.json
```diff
@@ -16,12 +16,18 @@
     "*.{js, jsx, css, scss}": "prettier --write"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^6.1.1",
```
jameskokoska, line 19: For icons, download the svg you want to use and import it
It reduces package count

client/src/components/Accordion/SingleAccordion/SingleAccordion.jsx
```diff
@@ -0,0 +1,47 @@
+import { React, useState, useRef, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faAngleDown } from '@fortawesome/free-solid-svg-icons';
+import './SingleAccordion.scss';
+
+const SingleAccordion = ({ header, children, toggle }) => {
```
jameskokoska, line 7: For initial value, pass in a prop `isInitiallyOpened`
and in the line below do  `const [isActive, setActive] = useState(isInitiallyOpened ? 'active' : 'closed');` for example
This initially opened might be needed for UX to show users that the accordion can be opened (by leaving one opened by default)

jameskokoska (changes requested):
Looks good though!

lukewarmtemp:
Removed the packages for font-awesome and downloaded the SVG for the arrow button. Made it such that the isOpen state and setIsOpen needs to passed from the parent function as props. Makes the code cleaner in the component and serves as a dual purpose for isInitallyOpen because we can pass isOpen = true/false.

jameskokoska (dismissed):
Looks good!

client/src/components/Accordion/SingleAccordion/SingleAccordion.jsx
```diff
@@ -0,0 +1,43 @@
...
+    setHeight(isOpen ? `${content.current.scrollHeight}px` : '0px');
+    setRotate(isOpen ? 'accordIcon rotate' : 'accordIcon');
+  }, [isOpen]);
+
+  function toggleAccordion() {
+    setIsOpen(!isOpen);
+  }
+
+  return (
+    <div className="accordian" onClick={toggleAccordion}>
```
Cali0707, line 22: fix the class name to be spelled properly

client/src/components/Accordion/SingleAccordion/SingleAccordion.scss
```diff
@@ -0,0 +1,31 @@
+@import '../../../scssStyles/mixins';
+
+.accordian {
```
Cali0707, line 3: fix classname to be spelled properly

Cali0707 (approved):
Looks good, there are two minor spelling mistakes, but I don't think those really matter.

PR #96, opened 2022-06-02
12 set up email api

Freeassassin:
Resolves https://github.com/UofT-Frosh-Orientation/orientation-website/issues/12.

The following functions were added:

`sendSimpleEmail(toAddresses, html, text, subject, fromAddress)`

A function which can receive a JS object containing either text or html and a list of email addresses and comprise and send the email(s).

`createTemplate(templateName, html, subject, text)`

A function to create and save an email template with fillable variables in the AWS dashboard.

`sendTemplateEmail(templateData, templateName, toAddresses, fromAddress)`

A function to send a templated email with filled in variables.

`sendBulkTemplateEmail(bulkEmailEntries, templateName, defaultTemplateData, fromAddress)`

A function to send templated emails with filled in variables to a wide range of addresses with different variables for each person.

`sendRawEmail(MIMEstring, toAddresses, fromAddress)`

A function to send a raw MIME formatted messages, which allows for the attachment of files onto the email.


server/src/services/EmailServices.js
```diff
@@ -0,0 +1,187 @@
...
+    };
+
+    const command = new CreateEmailTemplateCommand(params);
+
+    return SES.send(command);
+  },
+
+  /**
+   * Send a tempalted email with fillable variables
+   * @param {String} templateData data to be filled in the template example: "{ \"name\":\"Alejandro\", \"favoriteanimal\": \"zebra\" }"
```
Cali0707, line 95: This should be an object not a string, and then you convert it into a string inside the function

Cali0707 (approved):
Great work @Freeassassin !!

lukewarmtemp (approved):
LGTM!

PR #97, opened 2022-06-02
58 under maintenance page

natapokie:
maintenance page looks like this in desktop, 
image

and looks like this on mobile, 
image

the gear also spins!

let me know if the overall design is okay! i'm not sure how to do the navigation part though...

also removed some comments and testing files i used for the navbar. sorry, i forgot to check before!


client/src/pages/Maintenance/Maintenance.scss
```diff
@@ -0,0 +1,85 @@
+@import '../../scssStyles/mixins.scss';
+
+.full-page {
+  width: 100000vh;
+  height: 100000vh;
```
natapokie, line 5: this part looks very weird but the gradient didn't cover the full screen when i used 100% or 100vh...

jameskokoska: No problem, weird CSS sometimes does the job

client/src/pages/Maintenance/Maintenance.scss
```diff
@@ -0,0 +1,85 @@
...
+  align-items: center;
+}
+
+.gear-icon {
+  width: 150px;
+  height: 150px;
+  margin-bottom: 30px;
+
+  // colouring image white
+  filter: invert(100%) sepia(100%) saturate(0%) hue-rotate(160deg) brightness(105%) contrast(101%);
```
jameskokoska, line 39: An invert of 1 should do the trick, no need for other filters
Filters can be quite intensive for compute time

client/src/pages/Maintenance/Maintenance.scss
```diff
@@ -0,0 +1,85 @@
...
+
+  @include devices(mobile) {
+    width: 75px;
+    height: 75px;
+  }
+}
+
+// spinning animation
+@media (prefers-reduced-motion) {
```
jameskokoska, line 48: If it's reduced motion, shouldn't it not spin?
I think it can spin normally, just remove this media tag

client/src/pages/Maintenance/Maintenance.scss
```diff
@@ -0,0 +1,85 @@
...
+@keyframes icon-spin {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
+
+.title {
```
jameskokoska, line 62: title is a generic css class
since css is global in React use more specific class names, such as `maintenance-title`

client/src/pages/Maintenance/Maintenance.scss
```diff
@@ -0,0 +1,85 @@
...
+  font-weight: bold;
+  text-shadow: 0px 7px 10px rgba(0, 0, 0, 0.25);
+  margin-bottom: 10px;
+
+  @include devices(mobile) {
+    font-size: 20px;
+  }
+}
+
+.subtitle {
```
jameskokoska, line 76: Same, generic name

jameskokoska (changes requested):
I like the design and the animation is a neat touch!

client/src/pages/Maintenance/Maintenance.scss
```diff
@@ -0,0 +1,85 @@
...
+      rgba(255, 255, 255, 0) 0%,
+      rgba(221, 204, 231, 0.21875) 38.02%,
+      #621790 100%
+    ),
+    #935ebf;
+}
+
+.box {
+  position: fixed;
+  top: 50%;
```
jameskokoska, line 18: Line 18 to 22 might not be needed
`display: flex` should handle the centering of content with
```
justify-content: center; align-items: center;
```
what you used below

jameskokoska (dismissed):

client/src/pages/Maintenance/Maintenance.scss
natapokie: i'm not sure why, but it doesn't seem to work 😅

jameskokoska: Try removing `position: fixed`

jameskokoska: Check my latest commit

jameskokoska:
@natapokie go to `http://localhost:3000/maintenance`
It should be centered and I added the route as well
Storybook doesn't do the measurements justice

jameskokoska:
Also to be consistent could you name the component PageMaintenance

natapokie:
Will do! Thanks for fixing it James!

jameskokoska (approved):

client/src/pages/Maintenance/Maintenance.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+import GearIcon from '../../assets/misc/gear-solid.svg';
+
+const PageMaintenance = () => {
+  return (
+    <div className="maintenance-box">
+      <img className="gear-icon" src={GearIcon} alt="gear"></img>
+      <div className="maintenance-title">UNDER MAINTENANCE</div>
+      <div className="maintenance-subtitle">
+        We apologize for the inconvenience, we will be back soon! Check back here in a few minutes.
```
Cali0707, line 12:
```suggestion
        We apologize for the inconvenience, we will be back soon! Please check back here in a few minutes.
```

client/src/util/pages.jsx
```diff
@@ -48,5 +49,10 @@ export const pages = {
       component: <div />,
       path: '/permission request',
```
Cali0707, line 50: I am pretty sure this is an invalid path, can we change it to '/permission_request'?

Cali0707 (changes requested):

Cali0707 (approved):
This looks great, thanks @natapokie !

PR #101, opened 2022-06-04
fixing navbar icons!

natapokie:
i made a mistake with the icon sizing! it didn't show up in storyboard, but showed up on the beta version! 
the icons should look like this on mobile now!
image

instead of this....
image

sorry about that !!


jameskokoska (approved):

Cali0707 (approved):
Thanks for the fix - this looks great!

PR #114, opened 2022-06-07
a styles fixed!

natapokie:


jameskokoska (changes requested):
We want to avoid setting the CSS of all <a> tags because it will affect the styles of actual links
See http://localhost:6006/?path=/story/text--text-link
You can try using CSS inheritance selector, for example 
image

Also can I request you also change the name of the css classes used for the navbar
container, main, and special and too generic and should be something like navbar-main, navbar-container, navbar-special etc

Just because CSS is globally set. I should've requested changes for these names in the initial Navbar PR, my bad

jameskokoska (approved):
Great job thanks!

Cali0707 (approved):
👍

PR #123, opened 2022-06-09
Registration form and fields

jameskokoska:
Also modified some form components


jameskokoska:
Format of registration form needs work
And validation

it is a good start
I want to merge as I modified for components

PR #133, opened 2022-06-11
80 re design users

Cali0707:
This is a large PR that redesigns the way we handle users and adds all the routes required for registration. Please note, there is still some work to do in the issues this PR is resolving - I will extract that into new issues for the coming week.


server/src/components.yaml
```diff
@@ -25,50 +23,47 @@ components:
         discipline:
```
jameskokoska, line 23: So every user will need a discipline (before registering?)

jameskokoska (approved):
Looks good!

server/src/components.yaml
Cali0707: The code which we copied over from last year to generate the frosh groups requires that everyone has a discipline at the time they register, so currently yes. Do you think it is worth changing this?

jameskokoska: No, it's actually fine because it will make the actual registration have less steps
Which ones are currently needed on making an account?

By the looks of it is it
Email
Name
Password
Discipline
Pronouns

Cali0707: Just email, name, password.

Discipline and Pronouns are needed when registering

jameskokoska: Ohh i see I thought it was for just creating an account
Okay thats fine then

Freeassassin (approved):
Looks great!

PR #135, opened 2022-06-12
46 host initial landing page on orientation.skule.ca

Cali0707:
This PR makes it so that only the initial landing page is shown in production. The hosting on orientation.skule.ca is up to the webmaster to configure, but this will ensure that our prod deployment displays the initial landing page only for now.


jameskokoska (approved):
Great!

PR #143, opened 2022-06-14
103 about us page frontend (work in progress!)

natapokie:
update on about us page...
- [x] need to get exec images
- [x] need to fix mobile mode

looks like this so far...
image
image


jameskokoska:
Looks great!
Maybe reduce the area of the header image
You can do that buy modifying these lines (highlighted in the blue sidebar)

image

natapokie:
About us page almost done!
I used the photoshopped photos, but I want to see what it would look like with just the regular photos, so I am waiting for a reply from photography

On desktop...
image
image

On mobile...
image
image

- [ ] hopefully, i can get another picture of the execs for the header image
- [ ] need to delete all my comments after trying regular photos
- [ ] need to add exec writeups (will adjust textbox parts later if needed)

jameskokoska:
Looking good!!

jameskokoska:
I still don't approve of the About Us image at the top of the page

natapokie:
latest commit 67bff63 ...
- added all exec bios
- removed landing page/header image
- added fade in transition for the exec description popup
- added scrollbar - will be using something similar for the login page!

looks like this!
image
image
image

MatthewGWilson (approved):
I can't really speak for if the code is solid but I tried opening it and the visuals look great, clicking on the exec photos to see the description works well, and it overall it seems good to me! Great work Natalie!

jameskokoska:
Looks great! Nice animations and good use of setState

I changed the about us image, compressed images, changed the underline color, and fixed some formatting fixes when resizing the screen

jameskokoska:
This might come in handy
for css you can use min and max
It will take either the min or max of that measurements (for anything, not just width - where width has it's own max-width property)

`min(measurement, measurement)`
`max(measurement, measurement)`

E.g.
`margin: min(1vw, 10px)`
So this will take the margin of 1% of the screen width, until 1% of the screen width becomes 10px and it will min out at 10px and not be larger

Also,
 you can use these quotes: ` to have a neater string input (for very long strings, you can have new lines and it won't affect the output)
See: https://github.com/UofT-Frosh-Orientation/orientation-website/pull/143/files#diff-0952610a49efbad0bdda8843059f79f83e1f05a9305325469ad6255218f81340

natapokie:
thanks for the tips james!! also sorry i used your photo, i should've checked with you first 😅
also! I'll see if we have another frosh icon/graphic to use instead of the main logo, since the colour is the same, but also for some variation!

image

PR #154, opened 2022-06-15
Homepage slideshow, fixed carousel center

jameskokoska:


natapokie (approved):
looks good! added a small change to the carousel, to links would open in a new tab

PR #166, opened 2022-06-17
93 main schedule backend

### LINKED ISSUE #93: Main schedule backend
- [x] a route to read a list of Schedule objects by frosh group (no permissions) GET
- [x] a route to change/edit a Schedule objects (needs scope schedule:edit or schedule:all) PUT
- [x] a route to add a Schedule object (needs scope schedule:add or schedule:all) POST
- [x] a route to delete a Schedule object (needs scope schedule:delete or schedule:all) DELETE
- [x] a route to change the order of a Schedule object (needs scope schedule:edit or schedule:all) PUT

Freeassassin:
Resolves #93 

Added backend routes for the schedules


server/src/controllers/ScheduleController.js
```diff
@@ -0,0 +1,88 @@
...
+const ScheduleController = {
+  /**
+   * Get an array of schedule objects for a frosh group
+   * @param {Object} req
+   * @param {Object} res
+   * @param {Function} next
+   * @return {Object[]}
+   */
+  async getGroupSchedule(req, res, next) {
+    const { froshGroup } = req.body;
```
Cali0707, line 12: `req.body` should not be used in GET requests. In general, bodies should not be sent in a GET request, so trying to access them is an anti-pattern (if technically possible). What you can do is use the `req.user` object which the passportjs middleware is adding, which will give you access to the full mongoose document for the current user

server/src/controllers/ScheduleController.js
```diff
@@ -0,0 +1,88 @@
...
+
+    res.status(200).send({ groupSchedule: await ScheduleServices.getGroupSchedule(froshGroup) });
```
Cali0707, line 14: i don't like this pattern you've been using of calling a function directly within the `.send` method - too many chances for things to go wrong. Please refactor anything like this out into two distinct calls: one to get the result of the awaited function, and another to send it in the `res.send()`.

server/src/controllers/ScheduleController.js
```diff
@@ -0,0 +1,88 @@
...
+  },
+
+  /**
+   * add a schedule object
+   * @param {Object} req
+   * @param {Object} res
+   * @param {Function} next
+   * @return {Promise<void>}
+   */
+  async addSchedule(req, res, next) {
```
Cali0707, line 59: instead of adding a complete schedule, I would rather have an endpoint that allows someone to add an event to the schedule for the frosh group. This endpoint should have some logic that would create a schedule for said frosh group if one does not already exist (https://stackoverflow.com/a/33401897/14216419)

server/src/routes/scheduleRoutes.js
```diff
@@ -0,0 +1,188 @@
...
+ *     responses:
+ *       '200':
+ *         description: Schedule was successfully deleted
+ *         content:
+ *           application/json:
+ *             schema:
+ *                $ref: '#components/schemas/Schedule'
+ */
+router.delete(
+  '/delete',
```
Cali0707, line 182:
```suggestion
  '/delete/:id',
```
Again, for delete requests using the request body is an anti-pattern (google cloud will automatically reject the request, as an example). So, you can use a url parameter to specify which schedule to delete. However, I don't think it makes sense to have an endpoint to delete the entire schedule - you are going to want to have one which deletes a specific event from a schedule

Cali0707 (changes requested):
Overall, I really liked the way you structured your code - it is very well documented and clean! I left a lot of comments because I feel like there was some confusion on how exactly this was supposed to work. What you did does work, but it would be very difficult to work with going forwards as schedules are often edited quite frequently, and handling times is more complex than just rearranging indexes.

Also, there are many places where you update models by doing something along the lines of: 
```javascript
const schedule = await ScheduleModel.findById(id);
const newSchedule = ScheduleModel.hydrate(schedule.toObject());
newSchedule.set({ isDeleted: true });
return await newSchedule.save();
```
I'm guessing that this is coming out of you reading how I wrote the code where I updated the user model into a frosh model. In general, this is not required to update a model and is in fact worse than the normal way of handling things. If you look closely at what is happening in the code where I used it, the discriminator of the model was being updated, which is why I had to take a more unconventional approach. In general you can just do:
```javascript
const schedule = await ScheduleModel.findById(id);
schedule.set({ isDeleted: true });
return await schedule.save();
```

Freeassassin:
Changed the structure of the schedule to be an array of sub documents within the frosh group model. Each event doc contains a date and end time property that together allow for the display of the full time and date of an event. 

The reorder function was changed to be able to receive an array of objects that detail new start and end times for each event by id.

Also added requested changes regarding HTTP requests

jameskokoska:
Hi @Freeassassin, I hope it's not too painful to change/modify but could you follow the template given by @MatthewGWilson 
https://github.com/UofT-Frosh-Orientation/orientation-website/blob/dev/client/src/assets/schedule/data.jsx

Freeassassin:
Hey @jameskokoska, instead of changing the whole structure I made it so that the get request for the schedule will output in the format that is in the file you linked. 

If this is not good enough lmk and Ill make the changes.

jameskokoska:
Ya no worries that'll be good
I'll review tomorrow

jameskokoska (approved):
It looks good, but please wait for @Cali0707 to approve when he has the time

server/src/services/ScheduleServices.js
```diff
@@ -0,0 +1,186 @@
...
+      hours = hours % 12;
+      hours = hours ? hours : 12; // the hour '0' should be '12'
+      minutes = minutes < 10 ? '0' + minutes : minutes;
+      var strTime = hours + ':' + minutes + ' ' + ampm;
+      return strTime;
+    }
+
+    const group = await FroshGroup.findById(froshGroupId);
+    let schedule = {
+      'Monday September 5': [],
```
jameskokoska, line 23: Should this be hard coded?

server/src/controllers/ScheduleController.js
```diff
@@ -0,0 +1,102 @@
+const ScheduleServices = require('../services/ScheduleServices');
+
+const ScheduleController = {
+  /**
+   * Get an array of events for a frosh group.
+   * @param {Object} req
+   * @param {Object} res
+   * @param {Function} next
+   * @return {Object[]}
```
Cali0707, line 9:
```suggestion

```
This function does not return anything, so having a return value in the doc comment is misleading

server/src/controllers/ScheduleController.js
```diff
@@ -0,0 +1,102 @@
...
+   */
+  async getGroupSchedule(req, res, next) {
+    const { froshGroupId } = req.params;
+    const schedule = await ScheduleServices.getGroupSchedule(froshGroupId);
+    res.status(200).send({ groupSchedule: schedule });
```
Cali0707, line 14: Please wrap this in a try catch so that if there are any errors we handle them appropriately

server/src/components.yaml
```diff
@@ -67,6 +67,34 @@ components:
...
+          description: Name of the event
+          example: Matriculation
+        description:
+          type: string
+          description: Description of the event
+          example: lorem ipsum
+        date:
+          type: date
+          description: Date of the event including starting time
+          example: 2022-05-09T14:00:00
```
Cali0707, line 89: should be `startDateTime`

server/src/components.yaml
```diff
@@ -67,6 +67,34 @@ components:
...
+        endTime:
+          type: date
+          description: End time of the event
+          example: 2022-05-09T15:00:00
```
Cali0707, line 93: Should be `endDateTime`

server/src/services/ScheduleServices.js
Cali0707: No, this is not what I was picturing when I asked to separate the schedule and the events. I'll refactor this all to fix it

Cali0707 (changes requested):
Overall, this is a lot better than before! Two things (which I will be fixing due to time constraints):
1. try and use more descriptive variable names. To capture the start and end times of the events, you had `date` and `endTime`. From your comments, I can see that `date` also includes the start time of the event, but that is not immediately clear.
2. Don't hardcode the days of the week for the schedule - people may want to schedule things on other days and we do not want to have to come back and update this code to support that.

PR #169, opened 2022-06-18
faq backend routes (WIP)

NeoLou:


Cali0707:
I've updated this to have all necessary routes. Some notes:
- the PATCH `/faq/:faqId` route has an optional `faqId`, this is to make it easier to work with on the frontend where admins can add answers, as they can also just write a new question (which would not have an id yet), and post it and the new object would be created and stored
- We should probably add a route for admins to get all faqs, even the answered ones - I have not implemented that yet but there is a ticket for it

NeoLou (commented):
Looks good, but I was wondering if we want to keep the "updated" property so that we can sort the questions by most recently updated?

Also, why do we keep `{ strict: true }` in the model (isn't `strict` enabled by default)?

PR #176, opened 2022-06-20
FAQ-Front-End (WIP)

lukewarmtemp:
Things to implement:
- Interfacing with backend routes
- Popular and recently asked questions
- Category button selector formatting


Cali0707:
@lukewarmtemp what is the status of this?

lukewarmtemp:
> @lukewarmtemp what is the status of this?

Almost done

lukewarmtemp:
Desktop: Normal page at rest, search query used, question selected from search query
FAQ1
FAQ2
FAQ5

Mobile: Normal page at rest, search query used, question selected from search query
FAQ3
FAQ4
FAQ6

lukewarmtemp:
Mobile: Normal page at rest, search query used, question selected from search query

lukewarmtemp:
@jameskokoska I don't really know what to do with the button selector component right now, but right now it creates evenly sized buttons that fit to match the parent.

jameskokoska (changes requested):
- [x] Is the new wave image needed? Could you use the ones we already have and apply a shadow to it, similar to the home page
- [x] You should use Frosh purple ```var(--purple)```
- [x] On a larger display the space between the magnifying glass and clear button is quite large away from the sides, use px measurements to avoid sizing with the display
- [x] The dropdown question suggestion page should only have rounded corners on the bottom
- [x] Animate the hover effect for the suggested questions (similar to accordion component)
- [x] When clicking a suggested question from the dropdown, the textinput should also have that question in it, pre filled
- [x] Hitting enter in the search bar should search that query
- [x] When I select Question 1 in the search dropdown, it should select the tab that it's from (so e.g. General)
- [x] Clicking the search icon should also search for the question
- [x] `Can't Find Your Question?` text input box should use the text input component and follow those styles
- [ ] The user should get feedback after submitting the question, include a mock database call, similar to the one found in the sign-up page - it will load and provide the user with the result. use the <ErrorBox/> component to indicate success / error
- [x] The accordion should have pointer currsor when hovering
- [x] Make the header of the question accordion larger and bold
- [x] on wide screens the question box should be on the side, as shown in the design Figma
- [x] You should add an attribute that removes the stretching of the ButtonSelector to props
- [x] The shadow is a bit harsh on the dropdown search suggestions, use a softer shadow - from our color dictionary `box-shadow: var(--shadow)`
- [ ] cursor should be pointer when hovering over clickable actions, such as the clear text in the search input box

lukewarmtemp:
Implemented search that returns all questions which include the query string when the search icon is clicked or the enter key is pressed. Also implemented a 'no results' if search occurs with an empty string or no questions match the search query.

jameskokoska:
> - [ ] The user should get feedback after submitting the question, include a mock database call, similar to the one found in the sign-up page - it will load and provide the user with the result. use the component to indicate success / error

For this, use something similar tot he SignUp page
```js
export async function signUpUser(user) {
  console.log(signUpUser);
  /*eslint no-undef: 0*/
  let promise = new Promise((res, rej) => {
    setTimeout(() => res('An error occured!'), 1000);
  });
  let result = await promise;
  return true; //return an error message string to be dispayed, if an error
}
```

So you can implement a loading and success/error messages

If you want it to succeed, just return true from this function, instead of result
This promise just adds a delay in an async function, similar to requesting a response to a server and waiting until the credentials are valid/invalid
Here we wait 1000ms to simulate awaiting a response from the server backend api
Common practice, we should reject the promise since an error occurred, however in this simple use case I resolve the promise with the error message

client/src/pages/FAQ/FAQ.jsx
```diff
@@ -1,9 +1,471 @@
...
+  for (let i = 0; i < data.length; i++) {
+    if (data[i].category == 'General') {
+      generalQuestions.push({ question: data[i].question, answer: data[i].answer });
+    }
+    if (data[i].category == 'F!rosh Kits') {
+      FroshKitsQuestions.push({ question: data[i].question, answer: data[i].answer });
+    }
+    if (data[i].category == 'F!rosh Group') {
+      FroshGroupQuestions.push({ question: data[i].question, answer: data[i].answer });
+    }
```
Cali0707, line 36: Is there a reason why these are hardcoded? It seems strange to hard code three categories explicitly

jameskokoska:
I cleaned up the styles, but you still need to add the submit feature
Like you need to be able to handle error codes, clear the input if submitted, show a success message if the function returns true

client/src/pages/FAQ/FAQ.jsx
lukewarmtemp: Resolved by using an object and properties to dynamically add new categories based on the categories set in the data.

NeoLou:
image

When you hover in these areas the cursor changes to a pointer but it shouldn't right?

Also has this been done? -> Animate the hover effect for the suggested questions (similar to accordion component)

NeoLou (changes requested):
image

When you hover in these areas the cursor changes to a pointer but it shouldn't right?

Also has this been done? -> Animate the hover effect for the suggested questions (similar to accordion component)

lukewarmtemp:
> 

Hover mouse icon area fixed for accordion. In terms of the hover effect for the suggested questions, I think that it's done? I'm not sure if you wanted to accordions to have an animation as well @jameskokoska?

jameskokoska:
> Hover mouse icon area fixed for accordion. In terms of the hover effect for the suggested questions, I think that it's done? I'm not sure if you wanted to accordions to have an animation as well @jameskokoska?

Animations are good

NeoLou (approved):
LGTM!

jameskokoska:
- [ ] should show loading symbol when submit is waiting
- [ ] slightly increase horizontal padding of accordion to match that of search results (see images)
image
image
- [ ] when showing search results, the category tab bar should be 'unselected' or disappear entirely 
- [ ] Hitting the clear text button in the search bar, or searching for an empty string should exit the search mode

jameskokoska (changes requested):
See requested changes in comment above

lukewarmtemp:
@jameskokoska fixed things you listed and also implemented a thing where if the search bar is empty, search mode is off (so that if users delete everything in the search bar, the delete icon will disappear and therefore can't press it to get out of search mode)

jameskokoska:
Just a few more things
- [ ] make it so it searches automatically actually (it gives better feedback to the user)
- [ ] If you hit enter on an empty string, the page gets stuck in 'search' mode
- [ ] For the loading animation, can you just make the 'Can't find your question?' area float away (doesn't make sense for the whole page to fly away)

lukewarmtemp:
@jameskokoska
1. To implement the search as the user types feature, I just removed the option for the suggested question to appear on the search bar because it ends up being redundant if the page is already showing the matching results.
2. Relating to the previous point, the enter key and clicking the search icon button functions have been disabled.
3. Do we need an animation when switching in and out of search mode? I feel like the sudden switch is quite jarring.
4. I just commented out the code relating to the search bar suggested question, pressing enter/clicking the search icon for now. Do you want me to remove them or keep them as comments for the future in case we wanted to go back?

jameskokoska:
> @jameskokoska
> 
> 1. To implement the search as the user types feature, I just removed the option for the suggested question to appear on the search bar because it ends up being redundant if the page is already showing the matching results.
> 2. Relating to the previous point, the enter key and clicking the search icon button functions have been disabled.

Sounds good

> 3. Do we need an animation when switching in and out of search mode? I feel like the sudden switch is quite jarring.

This is a good idea

> 4. I just commented out the code relating to the search bar suggested question, pressing enter/clicking the search icon for now. Do you want me to remove them or keep them as comments for the future in case we wanted to go back?

Up to you, it will always be in commit history as well

jameskokoska:
- [ ] Animation switching between search and non search mode
- [ ] Remove the on click function and cursor pointer when clicking the 'search' icon (since search automatic)
- [ ] Can you add a floating button on this page that will scrollto the `Can't Find Your Question?` section (if there are a lot of questions, users might not scroll to the bottom to ask a question)
Something like: image

jameskokoska (changes requested):
See above

Cali0707:
@jameskokoska Do you think this is ready to merge?

jameskokoska:
I'll check it out tonight

jameskokoska:
- [ ] `inputType={'textArea'}` should be `text` for email
- [ ] The floating action button probably works better in the bottom right corner, with an icon (The button also doesn't stand out, maybe add a bigger shadow and change the color to darker purple or yellow?) Ask @natapokie for some inspiration if needed
- [ ] The ask your question textbox should have a max width, I can resize infinitely 
- [ ] Email placeholder can just  be `placeholder={'Email'}`
- [ ] Question placeholder can just be `placeholder={'Type your question here...'}`

jameskokoska:
Ok I fixed it up
There was a bug with empty space. Elements still take up space even when visibility set to hidden
You need to set display to none
By doing so, I removed the animation
You can try and add back in the animation if you want
Check my commits for other changes I made

You can also clean up the code as quite a bit is unused, such as displaying a single search question as a result

Messy/confusing code is okay (it's the nature of React, it will become quite messy. With practice it becomes easy to read), but unused messy code is confusing

The changes I requested earlier and the stuff I fixed now should not have slipped in to a review ready PR
Please do more thorough testing and more polished feature writing in the future

jameskokoska (approved):

PR #196, opened 2022-06-23
added payment error page

natapokie:
not as cool as james' 404 error page, but i tried to add in some small animations !
and just a general question, is it alright to use graphics from other sources...? (this one is from freepik!), if not let me know and i will recreate/copy it on illustrator

- [x] added button "Back to Payment" - currently links to a '/payment' page, need to link this to the right page later
- [x] now that i look at it the right padding seems off... will fix later!

looks like this... 
image

(had fun doing the animations so i will get back to the login page now...😅)


client/src/pages/PaymentError/PaymentError.scss
```diff
@@ -0,0 +1,129 @@
...
+}
+
+.payment-error-text-container-title {
+  font-size: 60px;
+  font-weight: bold;
+  padding: 10px 10px 25px 10px;
+  color: var(--purple);
+  width: fit-content;
+
+  transition: 500ms ease-in-out all;
```
jameskokoska: All shouldn't be used as it reduces the performance drastically 
See `Avoiding animation pollution` in
https://github.blog/2021-01-29-making-githubs-new-homepage-fast-and-performant/

jameskokoska: Seems like this transition isn't needed actually, but you can still take a look at that document if you wish

jameskokoska (approved):
I made these small changes in a commit above

jameskokoska:
Nice work! I really like the animations
I made it so the image if shifted up a bit since the image has more whitespace at the top, just for visual balance
I turned on the wiggle animation all the time because it's really cool!

jameskokoska:
I linked the back to payment to take back to registration, ideally it takes you back to the last tab, but we will see in the future (also because they shouldn't need to refill out the form if it fails)

The width is fixed, use 100% instead of 100vw 

As for using graphics from other sources, as long as it's royalty free with attribution it should be fine
We should link our sources though, maybe in the about page (for FontAwesome too)
See https://github.com/UofT-Frosh-Orientation/orientation-website/issues/197

natapokie:
> I linked the back to payment to take back to registration, ideally it takes you back to the last tab, but we will see in the future (also because they shouldn't need to refill out the form if it fails)
> 
> The width is fixed, use 100% instead of 100vw
> 
> As for using graphics from other sources, as long as it's royalty free with attribution it should be fine We should link our sources though, maybe in the about page (for FontAwesome too) See #197

thanks james!! I'll make a small section on the about us page to do all the attributions! 👍👍

client/src/pages/PaymentError/PaymentError.scss
```diff
@@ -0,0 +1,129 @@
+@import '../../scssStyles/mixins';
+
+.payment-error-page {
+  width: 100vw;
```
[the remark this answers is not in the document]
natapokie: got it! i'll keep that in mind for future pages too -- I'm pretty sure i used vw and vh for the login page which is why the popup didn't perfectly center!

PR #220, opened 2022-06-27
109 task and announcements backend

Emily9023:
backend for announcements

includes: 

- [x] View all posted tasks and announcements (all)
- [x] Post new task and announcement (subcom only)
- [x] Delete task and announcement (subcom only)
- [x] Complete task and announcement (specific user only)
- [x] Get list of completed tasks and announcements (specific user only)

### DIFF
--- a/server/src/models/UserModel.js
+++ b/server/src/models/UserModel.js
@@ -63,6 +63,16 @@ const UserSchema = new mongoose.Schema(
       required: true,
       default: true,
     },
+    completedAnnouncements: {
+      type: [
+        {
+          announcementID: {
+            type: String,
+          },
+        },
+      ],
+      default: [],
+    },
   },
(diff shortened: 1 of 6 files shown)


server/src/models/UserModel.js
```diff
@@ -63,6 +63,16 @@ const UserSchema = new mongoose.Schema(
...
       default: true,
     },
+    completedAnnouncements: {
+      type: [
+        {
+          announcementID: {
+            type: String,
+          },
+        },
+      ],
```
Freeassassin, line 73:
```suggestion
      type: [String],
```

Freeassassin (approved):
Looks amazing! Good job 👍

Cali0707 (approved):
This looks good for now, we may want to consider updating this in the future, but I think this is good for registration purposes

PR #234, opened 2022-06-30
92 timeline routes

Emily9023:
- [x] a route to read a list of Timeline Date objects (no permissions)
- [x] a route to change/edit a Timeline Date objects
- [x] a route to add a Timeline Date object
- [x] a route to delete a Timeline Date object
- [x] a route to change the order of a Timeline Date object (only certain accounts can do this) <- changed this to just sort the timeline by date


server/src/services/TimelineServices.js
```diff
@@ -0,0 +1,53 @@
...
+    return new Promise((resolve, reject) => {
+      TimelineModel.create(TimelineElement, (err, newTimeline) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve(newTimeline);
+        }
+      });
+    });
+  },
```
Freeassassin, line 38:
```suggestion
  async updateTimelineElement(id, date, name, description) {
    return new Promise((resolve, reject) => {
      TimelineModel.findOneAndUpdate({ _id: id }, {date, name, description}, (err, Timeline) => {
        if (err || !Timeline) {
          reject('UNABLE_TO_UPDATE_TIMELiNE');
        } else {
          resolve(Timeline);
        }
      });
    });
  },

  async saveNewTimelineElement(date, name, description) {
    return new Promise((resolve, reject) => {
      TimelineModel.create({date, name, description}, (err, newTimeline) => {
        if (err) {
          reject(err);
        } else {
          resolve(newTimeline);
        }
      });
    });
  },

```

Freeassassin (changes requested):
Honestly amazing work! Unfortunately, I couldn't get it to work on my machine ( empty documents and no changes with edit) but like it only took a couple of quick changes to get it to run, I commented the changes I made!

Also @Cali0707 told me to use a deleted flag instead of actually deleting stuff off of the database, so maybe also add that?

Cali0707:
@Freeassassin @Emily9023 don’t worry about using a deleted flag here - that is only really a thing for frosh data

Freeassassin (approved):
Looks great!

Cali0707:
@Emily9023 can you resolve the merge conflicts here?

server/src/controllers/TimelineController.js
```diff
@@ -0,0 +1,54 @@
...
+      await TimelineServices.saveNewTimelineElement(date, name, description);
+
+      return res.status(200).send({ message: 'Successfully added Timeline element!' });
+    } catch (e) {
+      next(e);
+    }
+  },
+
+  async deleteTimelineElement(req, res, next) {
+    const id = req.params.id;
```
Cali0707, line 43:
```suggestion
    const {id} = req.params;

```

Cali0707 (approved):
This is great work!  I added comments about three very small stylistic changes, but I am going to merge and approve this anyways. Think about these as just future things to do

PR #237, opened 2022-07-04
209 connect signup frontend to backend

### LINKED ISSUE #209: Connect signup frontend to backend
We need to connect the signup frontend to the backend.

The definition of done for this issue is:
- [ ] All parts of the signup page which should be connected to the backend are.
- [ ] A user is successfully created and saved in the database with the appropriate info from the UI
- [ ] The UI handles all responses (success, failure) from the API appropriately

Freeassassin:
Resolves #209 

connected the sign up page front end to the back end database.

- [x] All parts of the signup page which should be connected to the backend are.
- [x] A user is successfully created and saved in the database with the appropriate info from the UI
- [x] The UI handles all responses (success, failure) from the API appropriately


NeoLou (approved):
LGTM! Here are some screenshots of when I tested the signup:

image

image

client/src/pages/SignUp/functions.jsx
```diff
@@ -16,12 +16,36 @@ export function validatePasswordLength(password) {
...
-  let promise = new Promise((res, rej) => {
-    setTimeout(() => res('An error occured!'), 1000);
+import axios from "axios";
+
+
+export function signUpUser(user) {
+
+  const result = axios({
+    method: 'post',
+    url: 'http://localhost:5001/user/signup',
```
Cali0707, line 26: Please don't use hardcoded urls for any requests - these will only work for your local machine!

server/src/services/UserServices.js
```diff
@@ -4,7 +4,7 @@ const EmailValidator = require('email-validator');
 const UserModel = require('../models/UserModel');
 const newUserSubscription = require('../subscribers/newUserSubscription');
 
-const passwordValidator = /(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[^A-Za-z0-9])(?=.{8,})/;
+const passwordValidator = /(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.{8,})/;
```
Cali0707, line 7: Please revert this change to what it was previously - we do want to be checking for a password that has at least one special character, and this regex does not work for that. For example, `testADA123` matches this regex, but should not be a valid password in our system

server/src/services/UserServices.js
```diff
@@ -14,13 +14,12 @@ const UserServices = {
...
+  async validateUser(email, password, firstName, lastName) {
     const user = await UserModel.findOne({ email });
-    console.log(user);
     if (user) {
       throw new Error('DUPLICATE_EMAIL');
     }
-    if (name === '' || name === undefined || name === null) {
+    if (firstName === '' || firstName === undefined || firstName === null || lastName === '' || lastName === undefined || lastName === null ) {
       throw new Error('MISSING_NAME');
     }
```
Cali0707, line 24: In general, I'd prefer to not be doing this in our validation function as this will be caught as an error already when you try and create an entry in mongo (if it is set up to not accept empty strings for these properties). Essentially, this just becomes another piece of code we have to maintain whenever we make changes to the schema of our data.
```suggestion

```

server/src/models/UserModel.js
```diff
@@ -3,10 +3,18 @@ const getResponseObject = require('../util/getResponseObject');
...
 const UserSchema = new mongoose.Schema(
   {
-    name: {
+    firstName: {
       type: String,
       required: true,
     },
+    lastName: {
+      type: String,
+      required: true,
```
Cali0707, line 12: Try to follow this: https://mongoosejs.com/docs/validation.html to implement the validation you were doing in the validateUser function here

Cali0707 (changes requested):
There is a lot of good work here, thanks for doing it! I left comments about some changes to be made before this is ready to go. Let me know if you have any questions about those changes

Cali0707 (approved):
Amazing work @Freeassassin !

PR #240, opened 2022-07-05
171 create stripe routes and component

Cali0707:
This PR expanded a bit as I was working on it. It contains the code to connect the registration frontend to the backend, handle the payments, and allow users to login (but not sign up). You will need to update your .env file to test this, so please message me for details.

To test:
1. Create an account for yourself using the api
2. Login to that account
3. Go to the registration page
4. Use the test cards from (stripe)[https://stripe.com/docs/testing] to test if the payment works
5. Upon a successful payment you will be redirected back to the orientation site and it will say on the orientation site that an error occurred - that is the expected behaviour currently, as there is a separate ticket to fix that page.


server/src/services/PaymentServices.js
```diff
@@ -0,0 +1,71 @@
...
+        submit_type: 'pay',
+        billing_address_collection: 'auto',
+        line_items: [
+          {
+            price: process.env.STRIPE_TICKET_PRICE_ID,
+            quantity: 1,
+          },
+        ],
+        mode: 'payment',
+        success_url: 'http://localhost:3000/registration-success',
```
jameskokoska, line 61: I assume this will be changed? from localshost

server/src/services/FroshServices.js
```diff
@@ -31,11 +31,16 @@ const FroshServices = {
...
    * @return {Promise<Object>}
    */
-  async upgradeToFrosh(user, newInfo) {
+  async upgradeToFrosh(user, newInfo, paymentIntent) {
     const frosh = FroshModel.hydrate(user.toObject());
-    frosh.set({ ...newInfo, userType: 'frosh' });
+    frosh.set({
+      ...newInfo,
+      userType: 'frosh',
+      payments: [{ item: 'Orientation Ticket', paymentIntent, amountDue: 13000 }],
```
jameskokoska, line 42: does this handle the discount?

jameskokoska (approved):

server/src/services/FroshServices.js
Cali0707: No, I'll update this later for discounts

server/src/services/PaymentServices.js
Cali0707: Yeah, good catch!

PR #241, opened 2022-07-05
connect timeline frontend to backend

NeoLou:
so I can't get it to work but I might have a lead on what is causing the issue:

image

image

the format of the date in the mongo database might not be correctly parsed to the timeline component, especially the `dates.map` function, but i don't know exactly what is causing the issues (maybe when we format the date with `let formattedDate = new Date(date.date).toLocaleDateString(undefined, options);`


jameskokoska:
Yes, formatted date expects an already parsed date that's not in ISO format 

It needs to be changed if the format it expected is changed

jameskokoska:
Try creating a new date object based on the string retrieved from the backend to pass into the component

jameskokoska:
Please add `linkLabel` and `link` attributes to backend (from https://github.com/UofT-Frosh-Orientation/orientation-website/pull/246)

NeoLou:
so in my last commit i "fixed" the static link issue by using ```useAxios``` instead of a static link but it didn't work on my machine so idk how to fix it...

jameskokoska:
@Cali0707 see message above

Cali0707:
@jameskokoska @NeoLou at this point it's probably easier for me to just do this given the time constraints

Cali0707:
@jameskokoska if you run this in docker with the updated .env file I send then this should be working now

NeoLou:
I finished the final adjusting touches, @Cali0707 and @jameskokoska

Cali0707 (approved):
Thanks for the great work @NeoLou

Issue #245, opened 2022-07-05
Localhost keys missing for some registration fields


jameskokoska:
as pointed out by @Cali0707, some fields such as name aren't stored in localhost
they are missing the storageKey attribute in the registration form

PR #249, opened 2022-07-06
Form validation attribute

jameskokoska:
image
This also fixes https://github.com/UofT-Frosh-Orientation/orientation-website/issues/245
Also made the example placeholder UtorID follow the 8 character limit
Example:
```js
     validation: (value) => {
        if (
          value !== undefined &&
          value.split('-')[0] !== undefined &&
          parseInt(value.split('-')[0]) >= 1920 &&
          parseInt(value.split('-')[0]) <= 2020
        ) {
          return true;
        } else {
          return 'Please ensure your birthday is between 1920 and 2020';
        }
      },
```

Added `maxLength` and `isUtorID` props to `Textinput`
Fixed `isPhoneNumber` from going over the limit

### DIFF
--- a/client/src/components/input/TextInput/TextInput.jsx
+++ b/client/src/components/input/TextInput/TextInput.jsx
@@ -21,6 +21,8 @@ const TextInput = ({
   inputTitle,
   isPhoneNumber,
   isInstagram,
+  isUtorID,
+  maxLength,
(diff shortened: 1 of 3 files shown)


Cali0707 (approved):
LGTM 👍

PR #259, opened 2022-07-07
204 create accounts approval page

natapokie:
some bugs that affect the quality of the code, but the code still functions properly... but i'm just stuck 😓
- [x] upon refresh and hitting send (before checking any of the boxes), the printed array should be empty -- but instead contains objects, however the approve and deny fields are either set to false or undefined
- [x] authentication requests tab currently adds another object into the array, instead I'm trying to push it into the auth array that is already there...
  image

the two tabs look like this, 
- **verified circle**: indicates whether email used corresponds with the list of emails associated with leedur accounts
- **approve/deny** section: when approve box is clicked, approve is set to true, when the deny box is clicked deny is set to true
- **approve all...** button will light up all the approve check boxes, and set approve to true
- **send** button: calls a function that will send the array to the backend, currently it prints out the approve and deny status to the console
- page navigator on all accounts page: will display a certain number of results per page, you can specify through AccountsApproval.jsx, `numResultsDisplayedProp`, arrow navigators can also take you to different pages
- did not make this page for mobile, web only! 😅

image
image


natapokie:
all bugs were fixed in latest commit 74e9784, thanks james!!!

jameskokoska:
- [ ] When clicking approve all scopes, the button should change to what the button does. It should change to Deny all scopes, or something similar
- [ ] For the tab component, maybe pass in a prop so that the next and previous buttons are hidden (doesn't make sense to show them here)
- [ ] The numbers on the bottom of the All Accounts page indicating the page should have cursor pointer
- [ ] The save button should provide feedback if submitted successfully or not `<ErrorSuccessBox/>`
- [ ] Have fun, if you want animate the check and exit buttons, maybe a hover animation 
- [ ] I think it should default to the x button instead of nothing selected, because if nothing is selected, then they are not that type of account
- [ ] The second tab maybe account permissions as the label
- [ ] Dont forget to make the max accounts per page larger, instead of 3
- [ ] Approve all verified accounts button should indicate it only does this for the ones on the page
- [ ] Also there should be an indication changes were made and tell the user to not forget to save, or something like that. Usually it displays the info from the database and the user has to go into edit mode, so they know when they exit edit mode it saves with that action. But in this case the user might easily forget to save (maybe it should save after each action automatically? - without save button)
Looks good and very functional!

jameskokoska (changes requested):
See comment above, ping when changed

Cali0707:
@natapokie this is not crucial for registration, so @jameskokoska and I have decided to drop this page from the scope of the launch. Please keep working on it, and when it is ready we will release it (ideally within a week of registration going live). Thanks for putting in all the hard work on this PR!

natapokie:
> @natapokie this is not crucial for registration, so @jameskokoska and I have decided to drop this page from the scope of the launch. Please keep working on it, and when it is ready we will release it (ideally within a week of registration going live). Thanks for putting in all the hard work on this PR!

@Cali0707 sounds good! thanks for giving me more time to work on it !!

natapokie:
in latest commits, 
- added edit mode, disabled pointer events if you are not in edit mode (to prevent incorrect clicking)
- added "changes made" message
- added ErrorSuccessBox when Save button is clicked

@jameskokoska let me know if you find any other issues!

jameskokoska:
It looks good!
There's some issues, I have attached a video

There is some flickering (fighting for values) somewhere in the state

https://user-images.githubusercontent.com/50821962/178336219-cf8d8b42-9a25-4e7f-b560-317d41972c72.mp4


The don't forget to save changes doesn't show up sometimes

natapokie:
> It looks good! There's some issues, I have attached a video 
> There is some flickering (fighting for values) somewhere in the state
> The don't forget to save changes doesn't show up sometimes

@jameskokoska I think I fixed everything now! 👍
just one quick note, if you save a change in edit mode, and then enter edit mode again, and exit without saving, the checkboxes will go back to the original/very first approve/deny states since the backend is supposed to update the array/object with all the account info. i will make an issue for that shortly!

jameskokoska:
Unapprove verified accounts should set the exit button to red, right now it just unselects both
image
image

Looks good though! 
The bug you mentioned might not be a bad thing, when it 'loads' old data it should load from the server the updated data
This is usually what happens after saving. It loads from the server again to see if the changes got updated properly
But just make sure you await when saving data from the server

natapokie:
in latest commit 2f7f8eb made sure the unapprove buttons display deny checkbox!

jameskokoska:
When this is ready to review, remove the draft PR label

natapokie:
in latest commit, d8b1b0d, added async function into the functions.jsx file, 
i left the response as an empty string, since the front-end will display a message according to the result

Cali0707:
@natapokie can you resolve the merge conflicts on this PR?

natapokie:
> @natapokie can you resolve the merge conflicts on this PR?

@Cali0707 resolved conflicts!

PR #261, opened 2022-07-07
update frosh information route

Emily9023:
route for updating frosh information

fields correspond to the fields that are displayed in the register form (which is the same as the edit form)

also, I am unsure how to test for this code as there's a lot of different models working together. If you have any idea how to do unit testing or creating instances of fake data to test the code pls let me know!! It would be a lot of help


Cali0707 (approved):
Looks good @Emily9023

PR #262, opened 2022-07-08
217 signup confirmation email

### LINKED ISSUE #217: Signup confirmation email
When a user signs up for a new account, they should be sent a confirmation email. Copy for this email is to be provided by @MatthewGWilson 

The definition of done for this is:
- [x] When a new account is created, an email is sent through AWS
- [x] The correct personal details are filled into the email.

Please note, this should be done in the `subscribers` folder of the application, not directly from a controller.

Freeassassin:
Resolves #217 

Added sign up conformation email subscription.

- [x] When a new account is created, an email is sent through AWS
- [x] The correct personal details are filled into the email.

To test run the back end and send a post request to `user/signup` route with the following request body:
`{
	"email": "Your Email", 
	"password": "P@ssword123", 
	"name": "John Doe"
}`

You should then receive a sign up conformation email from tech@orientation.skule.ca


NeoLou (changes requested):
Looks good, but couple of things:
- The email is automatically sent to the junk mail
- The title of the email is "conformation" instead of "confirmation"
- The images at the beginning of the email and end of email don't load (even after I bring it from junk to inbox)
image

Freeassassin:
Thanks for the review. In response to your comments:

> * The email is automatically sent to the junk mail

Unfortunately we don't have any direct control over that.

> * The title of the email is "conformation" instead of "confirmation"

Whoops, spelling mistake. Its been fixed.

> * The images at the beginning of the email and end of email don't load (even after I bring it from junk to inbox)

That is due to some limitations with the Amazon email API, we will be taking care of it later.

> * There are no personal details filled in the email...?

This email template doesn't have any personalized fields.

Cali0707 (approved):
Great work @Freeassassin !

PR #282, opened 2022-07-12
profile schedule is the same as homepage schedule

natapokie:


jameskokoska (approved):
Looks good, but can it be tested?

natapokie:
> Looks good, but can it be tested?

Sorry I should’ve tested it too, I’ll do that right now!!

jameskokoska:
Not sure if it can be tested actually

natapokie:
> @natapokie Can you go to an old commit, where the login wasn't hooked up to the Profile page, and test this Just so it can be merged safely

yep! I'll do that !

natapokie:
in latest commit cac9630
- used the coloured accordion components according to the homepage, not sure if this looks good though so i can remove it too 😅 
- added canOpen prop to accordion
- disabled pointer events for accordions with no description

this is what the schedule component looks like in a previous commit (a4b3978), when the login page is not connected to the profile page, 
image

jameskokoska:
Great thank you Natalie!

PR #300, opened 2022-07-15
added logout button

natapokie:
on desktop, it's on the navbar, and on mobile it's on the profile page, i made the button open up a popup in case people click incorrectly, the popup looks like this...
image

i tested the profile page logout button on a previous commit, and the mobile version will look something like this...
logout button on homepage

@Cali0707  i'll leave a comment on the places that need to be connected !


client/src/components/Navbar/Navbar.jsx
```diff
@@ -36,79 +38,112 @@ const Navbar = ({ isLoggedIn, froshInitials, isRegistered }) => {
...
+        setTrigger={setShowLogoutPopup}
+        heading={'Are you sure you want to logout?'}
+        exitIcon={true}
+        blurBackground={false}
+      >
+        <Button
+          isSecondary={true}
+          label={'Logout'}
+          onClick={() => {
+            console.log('logging out');
```
natapokie, line 57: @Cali0707 this is the logout button for desktop !

jameskokoska:
Natalie, if you want, you can try to hook it up yourself
Do something similar to requestPasswordReset
In `UserController.js`
```js
/**
   * Logs out the currently authenticated user.
   * @param {Object} req
   * @param {Object} res
   * @param {Function} next
   * @return {Promise<void>}
   */
  async logout(req, res, next) {
    req.logout((err) => {
      if (err) {
        return next(err);
      } else {
        return res.status(200).send({ message: 'Successfully logged out!' });
      }
    });
  },
  ```
  It's just an async function

Cali0707:
@jameskokoska I had to do some messy merge conflict stuff on the navbar - can you verify I didn't break it?

jameskokoska:
I added a few things, looks good!

jameskokoska (approved):

PR #302, opened 2022-07-15
Connected faq ask questions frontend to backend

lukewarmtemp:
Suggested testing procedure for QA:

1. Don't create an account or login
2. Go to FAQ page
3. Click on "Ask a Question" button
4. Submit an email and question
5. You should receive an error with the response message of the error telling you to login

6. Create a new account
7. Login to your new account
8. Click on "Ask a Question" button
9. Submit an email and question
10. There should be a success message

11. go to the mongo express port at localhost:8081
12. go to orientation, faqs
13. You should see the new question and email you just made, set as the general category as default

### DIFF
--- a/client/src/pages/FAQ/FAQ.jsx
+++ b/client/src/pages/FAQ/FAQ.jsx
@@ -15,6 +15,8 @@ import LoadingAnimation from '../../components/misc/LoadingAnimation/LoadingAnim
 import QuestionMark from '../../../assets/icons/question-mark-solid.svg';
 
 import { PopupModal } from '../../components/popup/PopupModal';
+import { userSelector } from '../userSlice';
+import { useSelector } from 'react-redux';
 
 const PageFAQ = () => {
   const [activeIndex, setActiveIndex] = useState(0);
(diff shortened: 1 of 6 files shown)


jameskokoska:
Wait wait
A few things, maybe you misunderstood
You shouldn't need to create an account to ask a question, since you enter your email
Maybe if you're logged in, prefill the email with the logged in email users email
The question shouldn't be added
It should be added to an invisible list only the frosh relations subcoms can look at on a web page (hidden - later - please create an issue for this)
Then frosh relations posts the answer and question using a form page we build which posts the question (hidden - later - please create an issue for this)

jameskokoska (changes requested):
See above

lukewarmtemp:
@NeoLou Do you have time to fix the backend routes as per James' requested changes? I feel like you have a better understanding of what's going on, but if not I can do it.

Cali0707:
@lukewarmtemp is this in a working state now?

lukewarmtemp:
@jameskokoska 
1. Login check function removed on the backend by Neo, now you can ask questions even when you aren't logged in
2. Questions when added from the "Ask a Question" post have their answer set as an empty string by default. The route /faq/answered only returns questions with answers. Therefore, if this route is used on the functions.jsx file for FAQ to replace the static questions that are currently being displayed, they will not be returned and displayed to the frosh.
3. Pre-filling logged-in users with their email is little difficult to implement relating to how the state changes are being rendered with the useEffect hook. I might be missing something obvious, but I don't think it's high priority right now to implement

jameskokoska:
You can use the initialValue for textInput component to prefill the email

I think you should set a separate bool instead of making the answer a blank string

PR #326, opened 2022-07-16
adding bug report link into footer

natapokie:
on desktop, hovering over the message will "wiggle" the bug icon and highlight text in white
image

on mobile
image


jameskokoska:
@natapokie I put it in the bottom right corner on desktop and also made the grid more responsive for about us page
Technically you should use a `display:flex` instead of grid, but grid is better for perfect square containers, but flex is better for centering and wrapping

grid is like old flex box (but no need to change, just a heads up) (flex box is fairly new, grid is old)

PR #377, opened 2022-07-25
Add date when question was asked

NeoLou:
added dateAsked and lastUpdated to faq model

### DIFF
--- a/server/src/models/FaqModel.js
+++ b/server/src/models/FaqModel.js
@@ -23,18 +23,9 @@ const FaqSchema = new Schema(
       required: true,
       default: false,
     },
-    dateAsked: {
-      type: Date,
-      required: true,
-      default: new Date(),
-    },
-    lastUpdated: {
-      type: Date,
-      required: true,
-      default: new Date(),
-    },
   },
   { strict: true },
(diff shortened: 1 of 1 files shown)


lukewarmtemp (approved):
LGTM!

Cali0707 (changes requested):
@NeoLou please read here: https://mongoosejs.com/docs/timestamps.html

This is an easier way to implement this functionality because it will allow us to not need to remember to update the lastUpdated field

NeoLou:
I changed the model to include timestamps to know the created date and updated date according to this documentation: https://mongoosejs.com/docs/timestamps.html

server/src/models/FaqModel.js
```diff
@@ -25,6 +25,7 @@ const FaqSchema = new Schema(
     },
   },
   { strict: true },
+  { timestamps: true },
```
Cali0707, line 28:
```suggestion
  { strict: true, timestamps: true },
```
For the options for a schema, they need to all be in the same object

Cali0707 (changes requested):

Cali0707 (approved):

PR #378, opened 2022-07-25
118 subcom faq page

### LINKED ISSUE #118: FAQ interface for subcoms
Form so Subcoms can submit answers and view all of Frosh's questions
Protected route for special leader accounts only

lukewarmtemp:
Resolves #118 

Todo: 
- [x] Restrict page access to leaders
- [x] Fix dark mode theming 

Intended Features:
- Submit a new question, answer, and category using the form at the top of the page. The new question should appear in the answered list immediately. A pop up showing a successful update to the database should be seen after the submit button is pressed
- If you create a new question using the "Ask a Question" button and refresh, you should see the new question under the general category without an answer
- If you click on the "edit" button for a question, the question, answer, and category can be changed
- Pressing the save button in edit mode should exit out of edit mode and the question/answer/category should be updated with your changes immediately
- Toggling the "Is Answered" checkbox in edit mode and saving should move the question into the respective category immediately
- Updating the category on any question should either create a new category or move it to the list of questions under and existing category immediately
- Clicking the delete button should delete any question immediately
- When there are no answered/unanswered questions, "no results" text should be returned


lukewarmtemp:
@NeoLou Can you help link with backend please?

lukewarmtemp:
image

lukewarmtemp:
image

lukewarmtemp:
The check mark to change a question from being answered to unanswered in edit mode needs to be added back in. See the old question box.
image

lukewarmtemp:
Also need to fix this:
image

lukewarmtemp:
The delete button is not working now.

lukewarmtemp:
@NeoLou Do you mind fixing the things I listed above?

lukewarmtemp:
@jameskokoska Do you think we need to implement a search result page navigator (ie. page 1, 2, 3, etc of questions)? I'm thinking that if we have too many questions, it would be pretty annoying to scroll all the way down to find a question, but not sure how hard/easy it would be to implement.

NeoLou:
> The check mark to change a question from being answered to unanswered in edit mode needs to be added back in. See the old question box. image

I think we don't need that one anymore because I changed the backend to automatically update whether it's answered or not so that the user doesn't have to manually do it

lukewarmtemp:
@NeoLou mobile also needs to be fixed somehow. It might be better to switch back to my old design with the checkbox for answered and unanswered:
image

jameskokoska:
> @jameskokoska Do you think we need to implement a search result page navigator (ie. page 1, 2, 3, etc of questions)? I'm thinking that if we have too many questions, it would be pretty annoying to scroll all the way down to find a question, but not sure how hard/easy it would be to implement.

Not needed right now
You can create an issue as a future enhancement

NeoLou:
So the only problem left (I think) is that the time does not display correctly (it is 4 hours ahead of the actual time) but other than that everything else should be good to go

lukewarmtemp:
Technically a bug, but it only happens if you have a really long single word, which isn't very likely in the first place. I'm thinking that we don't need to bother trying to fix this:
image

lukewarmtemp:
Also, the unanswered question category doesn't get prefilled for some reason, but we have no idea why it's not working at the moment:
image

NeoLou (approved):
LGTM

jameskokoska:
> Technically a bug, but it only happens if you have a really long single word, which isn't very likely in the first place. I'm thinking that we don't need to bother trying to fix this: image

You can easily fix this with styles
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

jameskokoska (changes requested):
Please implement permissions
Only leader accounts with the respective scopes can access edit endpoints and view the page through the frontend

NeoLou:
So I have implemented permissions and fixed overflow wrap, but I'm not sure how to test permissions... @jameskokoska @Cali0707 any tips?

lukewarmtemp:
@NeoLou can you have a look at this? I think we need to account for the maximum width of the category box to determine the number of characters in a word before wrap-around occurs. 
image

jameskokoska:
I wouldn't say it's a global bug because button labels are usually static 

I would just do an inline and if the characters are past a certain amount pass in the word break as the style prop

lukewarmtemp:
Text wrap around fixed by specifying a minimum width and changing the CSS to inline.
image

lukewarmtemp:
@jameskokoska Auth scopes fixed and tested, so I think the PR is ready for review. I talked with Calum and he said it was okay to have the auth scope fix in this branch, but I can make a separate PR and branch if needs be.

jameskokoska:
Great thanks! So I can close this PR? https://github.com/UofT-Frosh-Orientation/orientation-website/pull/441

I'm reviewing it now, but thanks for testing it too!

lukewarmtemp:
> Great thanks! So I can close this PR? #441
> 
> I'm reviewing it now, but thanks for testing it too!

@jameskokoska You know what I'll just go to the branch and apply my fixes. I'll let you know once it's done.

lukewarmtemp:
@jameskokoska Okay #441 is updated with the fix now.

jameskokoska (approved):
Made some changes - there was a security issue where frosh could post answered questions (we don't want them to do that!)

jameskokoska:
Anyways looks great! The frontend is intuitive and well polished, thanks for doing this!

PR #379, opened 2022-07-25
Scunt start

jameskokoska:
image
image
image
image
image
image

This PR is still a WIP

TODO:
- [x] Judge panel read QR code
- [x] Category filters for missions list
- [x] Sorting filters for missions list


client/src/pages/ScuntJudgeForm/scuntTempData.jsx
```diff
@@ -0,0 +1,68 @@
+﻿export const list = [
+  {
+    id: '1',
```
Cali0707, line 3: For this list, can we switch `id` to be something else? (Anything else would be fine) - we will run into issues with this + mongoose which we are running on the backend

Cali0707 (changes requested):
Very fancy! Just one small change to make this easier to work with on the backend and then this is good to go

Cali0707 (approved):
LGTM 🤩

jameskokoska:
Let me get the judge qr scanning in and i'll merge after
thanks for review!

PR #387, opened 2022-07-26
Added secret leader account param

jameskokoska:
Tap the dino icon on the create account page 5 times
and check the leadur account
the froshObj leadur value gets set to true/false accordingly


natapokie (approved):
so cool 🤩🤩 !!

PR #388, opened 2022-07-26
Scope and permission request form for leadurs

jameskokoska:
image


natapokie (approved):
sin 9138129, minor style change to the dropdown component that I didn't realize when I did dark mode,

added background colour 
image
image

but otherwise looks good!!

PR #411, opened 2022-07-27
New dark mode implementation

jameskokoska:


jameskokoska:
Colors are now stored in `colors.jsx`
First color is light, second is dark
They can be the same if no change occurs when switching modes
To change modes use the provider and context
`import { DarkModeContext } from '../../util/DarkModeProvider';`
`const { darkMode, setDarkModeStatus } = useContext(DarkModeContext);`
```js
setDarkModeStatus(!darkMode);
```

PR #418, opened 2022-07-29
408 add leedurs to about us

natapokie:
leedurs tab looks like this, 
- ig icon, links to respective group's ig account,
- email icon opens email to respective frosh group

on desktop,
image

on mobile, the teams tab section is also scrollable
image


NeoLou (changes requested):
Looks good Natalie! A potential accessibility improvement could be to add the names of the frosh groups (alpha, beta, gamma, etc.) under their icons because some of those icons might not be easily recognizable to frosh so they might not know what group is represented by a certain icon.

natapokie:
> Looks good Natalie! A potential accessibility improvement could be to add the names of the frosh groups (alpha, beta, gamma, etc.) under their icons because some of those icons might not be easily recognizable to frosh so they might not know what group is represented by a certain icon. @natapokie

good point, thanks neo!! i don't even know what all the greek letters stand for 😂

added names in latest commit, cbcc38b !
image

NeoLou (approved):
LGTM!

jameskokoska (approved):
Looks good! Did some small changes

PR #422, opened 2022-07-30
413 fixed dark mode colors!

natapokie:
note: 
- only implemented darkmode for pages that the frosh will see
- other pages like accounts page and frosh table, i didn't make them compatible with darkmode!


jameskokoska (approved):
Looks good, thanks for going through it again!

PR #437, opened 2022-08-02
372 judge page

natapokie:
- [x] add a popup that says "click to reveal bribes"
- [ ] update info with all the current scunt leedurs (using last years data with placeholder image as a sample)
- [ ] tech team, think of bribes!!

looks like this, formatting/layout is exactly the same as execs on about us page
- f!rosh must click ~~all profiles~~ **three profiles** in order to view tech team bribes
- popup to remind f!rosh to click profiles and see bribes only shows up the first time you open the page
- popup, confetti and scrolls to techteam bubble once the secret judges (us) are revealed!
image

Note for testing...
-  make sure to set all highlighted fields to 0 (tech team will only show up once three different judges have been clicked on)
image
Edit: From James - you can also use incognito! It has a new localstorage every time


jameskokoska:
Hi Natalie, I really like it!
Just a couple of suggestions

If it's all the judges, frosh may never see ours - maybe if they reveal 3 judges we can show ours (because I think there are 20 judges this year!)

When it's revealed maybe show a popup that they have revealed the secret judge - or maybe scroll to the top and have tech team animate in (something to make it obvious)

Also you can create our tech team bribes (and use our subcom photo? or maybe one that photography took if it fits in the square), but I would add as our first point - Wink wink we made the website... we have unlimited bribe points...

Just some ideas, if you don't have time / don't want to no probs

natapokie:
> If it's all the judges, frosh may never see ours - maybe if they reveal 3 judges we can show ours (because I think there are 20 judges this year!)

Sounds good! I didn't realize there were going to be that many judges 😮!
And I like the ideas!! I'll have time to implement them, so no worries ;)))

natapokie:
in latest commit: d7bf29a,
- when clicking three different judge profiles, tech team is revealed
  - once tech team is revealed, there is a popup and confetti (popup disappears automatically after around 5000ms), and window scrolls to y: 400px where the tech team bubble is
- added photography photo and added description for tech team!

image

NeoLou (approved):
LGTM! But I feel like the bribes about buying dinner and buying a switch might be too much... (I would feel bad if they actually did it...)

jameskokoska (changes requested):
- [ ] Extra margin at top of page (see screenshot ) 
image
- [ ] secret judges revealed shows up every time the page is loaded (in incognito) 
- [ ] For scrolling, you can just use 0 px is fine (or scroll to element) 400px might be different depending on the flex of elements above

jameskokoska:
Also @natapokie to keep track of which ones are clicked you should use only one local storage variable
You can keep track in an array and append the judge name if not already there and count the length
Since local storage can only store strings use something like:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify to convert between

jameskokoska (approved):
Looks great thanks!

PR #460, opened 2022-08-06
Create Timeline Admin page to edit timeline events

NeoLou:
This is a page (`localhost:3000/timeline-admin`) so that users with the right permissions (aka `authScopes :'timeline:edit'`) can access the database to perform the following actions:
- [x] Create new timeline events
- [x] Edit existing timeline events
- [x] Delete existing timeline events

Testing:
- Test whether these 3 functionalities work and if the display on the home page and on the admin page looks good


NeoLou:
@lukewarmtemp if you could test the functions and also figure out why the get function doesnt refresh without reloading (for example, after deleting an event it doesn't disappear without reloading the page)

NeoLou:
ok so latest update: everything should work now except that the existing events don't display properly after creating a new event or deleting an event

Cali0707:
@lukewarmtemp can you make sure that all existing events still work with this new PR? I suspect the issues are due to the mongo model change, so please make sure that any changes to that are backwards compatible

NeoLou:
@lukewarmtemp @jameskokoska I should have fixed all the issues so if you could review this PR when you guys are free that would be great!

@Cali0707 I tried adding swagger comments for the backend timeline routes but not sure if they're correct so could you double check that when you will have the time?

jameskokoska:
image
If link is blank or link text is blank there should be no button
The title doesn't appear on the popup window at the top

image
The title does not appear in the timeline

image
The faint text should be an example value (to show ppl how to use links for example)
For link, use something like https://www.google.ca

jameskokoska (changes requested):
See above

jameskokoska:
@lukewarmtemp Could you fix these changes, they seem quite simple
Since @NeoLou is working on another ticket atm

NeoLou:
@jameskokoska sorry for the delay, I made the changes
@lukewarmtemp no need to cover for me anymore lol

client/src/pages/Profile/Profile.jsx
```diff
@@ -251,6 +251,11 @@ const ProfilePageLeaderPermissionDashboardLinks = () => {
         authScopes={['faq:delete', 'faq:edit']}
         label="FAQ Admin Panel"
       />
+      <ProfilePageDashboardLink
+        link="/timeline-admin"
+        authScopes={['timeline:edit']}
```
natapokie, line 256:
```suggestion
        authScopes={['timeline:edit', 'timeline:create', 'timeline:delete']}
```

there are options for create, edit, and delete for timeline... on the leedur permission page `/permission-request` (?)
image

so it might be good to include all of them (as in suggestion)? or alternatively we can change the timeline scopes in the file (client > src > pages ? ScopeRequest > functions.jsx > getTotalScopes) to only edit?

natapokie (changes requested):
the timeline admin page looks great neo! the create, edit and delete buttons work as expected!!

just a few suggestions,
- would it be possible to add in a default Link Label (maybe `'Click Here!'`) if the label field is empty?
- would it be possible to show the button only when the Link field is not an empty? I can add a link Label but no link, but the button shows up (see below)
  image
  image

other than that, I think it's good to go !!! 🤩🤩

client/src/pages/Profile/Profile.jsx
Cali0707: The desired behaviour here would be that `timeline:edit` allows you to edit the contents of a timeline event, `timeline:create` allows you to create a timeline event, and `timeline:delete` allows you to delete a timeline event. Good catch @natapokie !

NeoLou: Got it, made the change!

NeoLou:
@Cali0707 @natapokie @jameskokoska I made the changes requested, please re review when you are free!

natapokie (changes requested):
the timeline looks great! thanks for making those changes neo!!

i'm just having a one small problems with the authscopes, 
if i only have `timeline:create` or only `timeline:delete`, the timeline admin panel button appears on the profile but I when I click it, i'll navigate back to the homepage 🤔

all the other scopes work though and give error messages when pressing buttons they don't have scopes for!

and one last thing if it's an easy fix, would it be possible to give an error message something like this, 'You don't have permissions to access this' or something similar, so leedurs know that they need to request more permissions? if it takes a long time, don't worry about it though 😅

NeoLou:
> the timeline looks great! thanks for making those changes neo!!
> 
> i'm just having a one small problems with the authscopes, if i only have `timeline:create` or only `timeline:delete`, the timeline admin panel button appears on the profile but I when I click it, i'll navigate back to the homepage 🤔
> 
> all the other scopes work though and give error messages when pressing buttons they don't have scopes for!
> 
> and one last thing if it's an easy fix, would it be possible to give an error message something like this, 'You don't have permissions to access this' or something similar, so leedurs know that they need to request more permissions? if it takes a long time, don't worry about it though 😅

Hmm so I think it would be a harder change because you would have to know what the error is as there can be multiple types of error...Maybe as an improvement for next week?

natapokie (approved):
looks good neo !!

Cali0707 (approved):
LGTM

PR #464, opened 2022-08-09
442 connect announcements to frontend

Freeassassin:
Connected announcements profile front end to the announcement backend

- [x] added redux infrastructure to front end
- [x] connect redux to profile page
- [x] add complete functionality
- [x] removed duplicate completed announcements


lukewarmtemp:
Announcements on profile page work great! I'm able to delete, edit, and send to top without any problems still.

Suggestions:
1. Complete entry functions works great, but do we need to implement an "undo" complete entry function? There might be a case where an announcement is marked completed by accident and needs to be marked as incomplete.

lukewarmtemp (changes requested):
See above

Freeassassin:
> 1. Complete entry functions works great, but do we need to implement an "undo" complete entry function? There might be a case where an announcement is marked completed by accident and needs to be marked as incomplete.

This is a valid suggestion but like the announcements do go away, they just go to the bottom of the list so like if its an accident the information is not lost.

lukewarmtemp (approved):
LGTM!

jameskokoska (approved):

PR #468, opened 2022-08-10
410 announcements dashboard

Freeassassin:
Created a new page `/announcement-dashboard`. 

- [x] edit 
- [x] delete
- [x] create
- [x] send to top
- [x] add authscopes


lukewarmtemp:
Not sure exactly what this arrow does, but it prompts the snackbar "Successfully Edited!". Don't know if it's supposed to do that or not.
image

lukewarmtemp:
LGTM! I was able to test and verify the edit, create, delete, send to top, and add auth scopes features.

A few changes that might need to be made:
1. Submitting a new announcement without either a name or description returns a snackbar error with 'Please provide your full name.' which might need to be changed to something else.
2. I get this problem where if a single word name and description (as in, without any spaces) of a newly created announcement are too long, it doesn't wrap around and the buttons to send to top, edit, and delete are out of frame. While it's unlikely that we will have a single word that long, it should be simple css fix.
image

lukewarmtemp (changes requested):
See above

lukewarmtemp:
Can you also apply the wrap-around CSS to the announcement name?
image

lukewarmtemp (changes requested):
See above.

lukewarmtemp (approved):
LGTM!

jameskokoska (changes requested):
@Freeassassin 

Looks good, well done!!
If you could make it so the user doesn't need to refresh when an announcement is created (maybe run the get request every time the tab is clicked) that would be appreciated!

Below has been completed in: https://github.com/UofT-Frosh-Orientation/orientation-website/pull/464

~~Could you also hook up marking off the announcement/task as completed~~

~~There's an entry in the user model where you can add the id of the announcement to an array
If a frosh has completed the task/announcement move it to the bottom of the list on the profile page~~

jameskokoska (approved):
Looks great thank you!

PR #470, opened 2022-08-10
added frosh group specefic viewing of registration data

jameskokoska:


natapokie (approved):
looks good and works !!

PR #479, opened 2022-08-13
Frosh Retreat frontend

jameskokoska:
image
image
image
image

The video will be added in later - if possible YouTube embed is the simplest


natapokie (approved):
looks good!! added a small style change for light mode,

image

PR #483, opened 2022-08-13
fixed style for reset password

natapokie:
changed box and header color, 
image
image

also fixed another small bug where it would display `undefinedundefined` on profile icon, after reseting password --> now shows profile icon as in above screenshot
image


Cali0707 (approved):
Looks great, thanks @natapokie !

jameskokoska:
Thanks for fixing this!

PR #485, opened 2022-08-15
Payment success and fail pages for retreat

jameskokoska:
/registration-success-retreat
and 
/payment-error-retreat


natapokie (approved):
looks good!!

PR #496, opened 2022-08-17
443 look up user and sign in qr scanner search

Emily9023:
mark isPresent flag as true when submit button is clicked
display frosh information once submitted

Questions:
- do we want to display all frosh, even those that arent registered??
- how do I search by name rather than firstname, lastname


jameskokoska:
Did a quick review:
Can you limit the search results returned to about 5 users
And yea only those who have `isRegistered` set to true is preferred 

searching by email only is okay actually, can you just update the frontend so it says Email only instead of Frosh Name / Email
image

jameskokoska (changes requested):
See above message

jameskokoska:
Also I would use dash notation for backend routes instead of camel case

jameskokoska (changes requested):
Not sure if the search is working correctly
image

NeoLou:
The QR scan doesn't work for me... I scanned with my phone and the frosh isPresent status is still false... I also don't know where the search bar is to test it

NeoLou:
Nevermind, tested QR code again and it works now, scanned and frosh was marked (isPresent changed to true). However, the feedback from a successful submission isn't great, as the container does become green but there is an additional snackbar with something like "Frosh successfully marked present!" would be good.

Also, the search function does not work for me (nothing shows up no matter what I type in).

NeoLou (dismissed):
See above

Freeassassin (dismissed):
Not sure what proper testing steps are but for me neither the QR scanner or the searched worked. The scanner would pull up the information but every time I pressed submit it gave me an axios 500 error and in the backend its throwing a `UNABLE_TO_UPDATE_FROSH` error. Also manual search didnt work for me at all.

natapokie (approved):
looks good!!!

- qr scanner manual displays up to five results,
- scanner submit button works and correspond to the sign in [date] parameter,
- authscope to view frosh sign-in dates works

Cali0707 (approved):

PR #512, opened 2022-08-28
add uncomplete announcement functionality

### LINKED ISSUE #503: Mark announcement as incomplete
When clicking the green check mark it will mark the task as incomplete (i.e. remove it from the list of completed announcements)

Freeassassin:
Resolves #503 

To test create announcement in `/announcement-dashboard` with leadur account. then create a frosh account and click the complete task button and then press it again to remove it from completed section.

### DIFF
--- a/server/src/services/AnnouncementServices.js
+++ b/server/src/services/AnnouncementServices.js
@@ -8,19 +8,9 @@ const AnnouncementServices = {
         if (err) {
           reject(err);
         } else {
-          resolve(
-            announcements.sort((a, b) => {
-              if (a.dateCreated > b.dateCreated) {
-                return -1;
-              } else if (a.dateCreated < b.dateCreated) {
-                return 1;
-              } else {
-                return 0;
-              }
-            }),
-          );
+          resolve(announcements);
         }
-      });
+      }).sort({ dateCreated: -1 });
     });
   },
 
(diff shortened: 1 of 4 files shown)


natapokie (approved):
looks good!!
i added in some imports since some functions were undefined 👍

server/src/services/AnnouncementServices.js
```diff
@@ -32,27 +22,45 @@ const AnnouncementServices = {
...
+                { completedAnnouncements: listOfCompleted },
+                (err, user) => {
+                  if (err || !user) {
+                    reject('UNABLE_TO_UPDATE_USER');
+                  } else {
+                    resolve(user);
+                  }
+                },
+              ),
+            );
```
Cali0707, line 62: This should be refactored so that only the `listOfCompleted.push` and `listOfCompleted.splice` operations are contained in the if-else block. The rest of the logic seems to be the same between the two, so it should be separated out.

Also, you should only ever call `resolve()` once in a Promise, as that is when the function will return - here your Promise is resolving to a new function which is then calling the old `resolve()` callback when it finishes. So, instead of `resolve(UserModel. ...)`, you only need to call `resolve()` inside the callback on the `UserModel.findOneAndUpdate()` method

Cali0707 (changes requested):

Cali0707 (approved):

PR #537, opened 2022-09-01
504 store title of announcement in backend

### LINKED ISSUE #504: Store title of announcement in completedAnnouncements as well
This might be useful if the leaders need to extract data from which tasks the frosh have not completed (instead of just the announcement id)

Freeassassin:
Resolves #504 

To test complete an announcement as a user. In the database the name and ID of the announcement should be stores in the user document.


Emily9023 (approved):
Looks good Farbod! Just a small thing, but the announcement list lags a little when I click the check mark. Not a huge problem but something I noticed. Great work tho :))

Cali0707 (approved):

PR #606, opened 2023-05-27
Landing Page - Sherry

Sherryzh0830:
## **Contribution:**
A testing page that gives a rough landing for work in progress.

### **Issue:** 

Closes # 598

### **Accomplishments:**
Was able to explore quite a few styling options and tried to make consistent with the existing files. 

### **Expected Behavior:**
- Have a message to let users know F!rosh is approaching.
- Display logo of F!rosh and able to navigate user to the social media page

### **Complications:**
Renamed and changed font, colors to be consistent to App.scss and unique to the tasks.


client/src/pages/Initial/LandingPage<Sherry>/LandingPage<Sherry>.scss
```diff
@@ -0,0 +1,37 @@
+* {
```
natapokie: when you're working on any styles, make sure to use very unique classnames, for example `landing-page-sz-top`.
in react, all the classnames and their style are shared with all components, so by making very unique classnames you'll be able to ensure that the style you have is not going to affect any components.

additionally, you can take a look at `App.scss`, this is where we define the style for "global" components like `*`, `:root`, `body`. you'll need to rename these in your component and use `className` to define these styles instead.

client/src/pages/Initial/LandingPage<Sherry>/LandingPage<Sherry>.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+body {
+  margin: 0;
+}
+.wrapper {
+  display: grid;
+  place-content: center;
+  background-color: var(--background-color);
+  min-height: 100vh;
+  font-family: 'Oswald', sans-serif;
```
natapokie: font-family is also already defined, we're going to stick with the current fonts that we have set up which is `Inter` you can find it in `App.scss` to ensure that we stick with branding guidelines.

since it's already defined "globally" there is no need to add font classes

natapokie (changes requested):
thanks for sharing your progress so far Sherry! I left some comments that you can take a look at as you complete the issue

also just a quick note, please make sure to rename your files so that they don't contain "<>" and any other special characters! these characters aren't very git friendly haha

the details in the issue weren't too clear, but the angle brackets, `<YourName>` were meant to indicate that you would fill in to replace the angle brackets like this... `LandingPageSherry`.
sorry for the confusion!

uzFer (commented):
My screen looks like this, are there any instructions I'm missing when running your landing page (I just imported and used it in App.jsx)?

image

uzFer (approved):
Works perfectly with the links and visuals, looks really nice!!

natapokie (approved):
great work sherry!

for the github file, you actually deleted it, but no worries! i made another commit to this branch to revert the changes, in the future, if you want to revert a file back to a previous commit you can use the following...

```
git checkout [commit ID] -- path/to/file
```
(no square brackets though!)
and then commit as usual!

more git reverting stuff [here](https://dev.to/lofiandcode/git-and-github-how-to-revert-a-single-file-dha#:~:text=look%20like%20this%3A-,git%20checkout%20%5Bcommit%20ID%5D%20%2D%2D%20path/to/file,-If%20I%20were)

PR #613, opened 2023-06-01
About Us Page Optimizations

### LINKED ISSUE #601: About Us Page
## **Changes:**

Make a placeholder on the about us page where it currently introduces all the team members -- something like "Stay tuned to meet to team!"
- Make sure to wrap the component `About.jsx > AboutUsTeamsTab` in another component (kinda  like a helper function) and use a state variable so we can easily hide/show the team
- Also add another property to the objects in the `tabs` array, so that it can show/hide specific teams (i.e., we only want to display the exec team)

In the `ExecProfile` component, introduce a timeout, so that after you click on the exec profile and it flips to the description, after a certain time it will flip back to the front photo.
- This feature is implemented because it gets annoying to purposely click back the component to view the photo again haha

## **Visuals:**

Any visuals that are related to these changes

## **Requirements:**

- [x] Placeholder for the entire team introductions
- [x] "State wrapper" to easily show all team introductions and team tabs
- [x] Introduce timeout to the ExecProfile component

## **Related Issues/PRs:**

Note, we'll need to gather all the info for the team later

## **Resources:**

### **Documentation:**

- [pages/About](https://github.com/UofT-Frosh-Orientation/orientation-website/tree/dev/client/src/pages/About)

vproHacks:
### **Issue:**

Closes #601 

### **Accomplishments:**

- Created a timeout on the `ExecProfile` component with a constant `EXEC_TIMEOUT` to configure how long it takes before defaulting
- Created a wrapper for the `AboutUsTeamTabs` component that now has the state variable `showAboutUs` to display or not (functionality shown with a simple toggle button)

### **Visuals:** 

Just used a default button

### **Expected Behavior:**

- Upon clicking on a profile it should revert back to the photo in 5 seconds
- On clicking the toggle it should toggle displaying the team alongside maintaining the current tab.

### **Testing Steps:**

- start the front end (no docker needed)
- Go to About page and scroll down to see the components

### **Complications:**

- Just getting used to Hooks and attempting to set timeout accurately in React

### **Resources:**

- [Help with react timers](https://felixgerschau.com/react-hooks-settimeout/)

### DIFF
--- a/client/src/pages/About/About.jsx
+++ b/client/src/pages/About/About.jsx
@@ -18,12 +18,14 @@ import InstagramIcon from '../../assets/social/instagram-brands.svg';
 import MailIcon from '../../assets/social/envelope-solid.svg';
 import { instagramAccounts } from '../../util/instagramAccounts';
 
+import { Button } from '../../components/button/Button/Button';
+
 const PageAbout = () => {
   return (
     <>
       <div className="aboutus-page-components">
         <AboutUsSection />
-        <AboutUsTeamsTab />
+        <AboutUsTeamsTabWrapper />
         <div className="about-attribution-container">
           <p className="about-attribution-message">
             Thank you to{' '}
@@ -284,18 +286,22 @@ const tabs = [
   {
     title: 'Exec Team',
     component: <AboutUsExecTeam />,
+    active: true,
   },
   {
     title: 'Tech Team',
     component: <AboutUsTechTeam />,
+    active: true,
   },
   {
     title: 'Subcoms',
     component: <AboutUsSubcom />,
+    active: true,
   },
   {
     title: 'Head Leedurs',
     component: <AboutUsHL />,
+    active: true,
   },
 ];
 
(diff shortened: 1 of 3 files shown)


vproHacks:
Added the functionality for placeholder, ready for PR review.

Tanushanth (approved):
Tested out the timeout, times out after 5 seconds properly! Also, I tried making the showAboutUs state false to see how it looks and it works perfect :).

client/src/pages/About/ExecProfile/ExecProfile.jsx
```diff
@@ -6,7 +6,10 @@ import 'react-lazy-load-image-component/src/effects/blur.css';
 import './ExecProfile.scss';
 
 import wave from '../../../assets/about/wave-about.svg';
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
+
+// Set 5s to read the post before closing automatically, configure as needed
+const EXEC_TIMEOUT = 5000;
```
natapokie: nice use of the timeout!
for now lets set the timeout to 25s (or maybe 30s?), I know it's a little long but reading some of the exec profiles does take a while ahha

client/src/pages/About/About.jsx
```diff
@@ -306,56 +312,88 @@ const AboutUsTeamsTab = () => {
...
-      </div>
+const AboutUsTeamsTabWrapper = () => {
+  const [showAboutUs, setShowAboutUs] = useState(true);
+
+  if (!showAboutUs) tabs.map((tab) => (tab.active = false));
+  else tabs.map((tab) => (tab.active = true));
+
+  return (
+    <>
+      <AboutUsTeamsTab />
```
natapokie: would recommend making use of the state here to check whether to display
`<h2 className="about-introduction-title">Stay tuned to meet the team!</h2>`
this is just makes the code a little cleaner and easier to follow since we're either displaying the tab component or the stay tuned text...

another note is that you can simply just use a variable `const showAboutUs` rather than a state, although both work!

you can do something like this with the use of a ternary operator...

```
return (
  <>
    {
      showAboutUs ?
      <Tab.. />
      :
      <Text />
    }
  </>
);
```

client/src/pages/About/About.jsx
```diff
@@ -306,56 +312,88 @@ const AboutUsTeamsTab = () => {
...
-        {tabs.map((tab) => {
-          if (currentTab === tab.title) {
-            return tab.component;
-          }
-        })}
-      </div>
+const AboutUsTeamsTabWrapper = () => {
+  const [showAboutUs, setShowAboutUs] = useState(true);
+
+  if (!showAboutUs) tabs.map((tab) => (tab.active = false));
```
natapokie: you might want to remove this `if...else`, since it sets all the active fields to either true or false, we won't be able to hide/show specific tabs

client/src/pages/About/About.jsx
```diff
@@ -306,56 +312,88 @@ const AboutUsTeamsTab = () => {
...
+            }
           })}
         </div>
-      </div>
+      </>
+    );
+  else
+    return (
+      <>
+        <h2 className="about-introduction-title">Stay tuned to meet the team!</h2>
```
natapokie: mentioned in the comment below, but this line is best returned in the `AboutUsTeamsTabWrapper`

client/src/pages/About/About.jsx
```diff
@@ -306,56 +312,88 @@ const AboutUsTeamsTab = () => {
...
-                    <h1
-                      className={
-                        currentTab === tab.title
-                          ? 'aboutus-teams-tabs-title-selected'
-                          : 'aboutus-teams-tabs-title'
-                      }
-                    >
-                      {tab.title}
-                    </h1>
+  if (tabs.every((value, index, array) => value.active))
```
natapokie: you can remove this statement here... `if (tabs.every((value, index, array) => value.active))`, you're checking for `active` in the actual component, so I'm not entirely sure what this line does?

and I didn't realize until now, but you'll need to make adjustments to the state to `setCurrentTab` to the first tab that is active! you can use a `useEffect`, with an empty dependency array, `[]` so it executes on first render

natapokie (changes requested):
great work vraj! really nice use of the timeout, works great!

just a quick note on the overall logic of the code, my explanation probably wasn't clear and way the code runs is different than what I intended when I first wrote the issue...

- we want to use the "wrapper" so that it either displays the tab component or the "stay tuned" text, which is why I suggest using the ? operator in the `AboutUsTeamsTabWrapper`
- we want to display **certain** tabs depending on the active state (this is probably where I was being unclear). 
  - the main reason for this is that we might want to show only the execs and nothing else, or maybe we would want to show the execs and hls and no one else. 
  - we can easily comment out the code, but we want it to look a little cleaner and intentional, so we're introducing this new `active` state.
  - so there is no need to "update" the `tabs` array by setting active to true in the `AboutUsTeamsTabWrapper`, this changes the purpose behind the `active` field in the object
  
hope this makes more sense, I might also be testing your code wrong as well, I just couldn't find a way to only show one tab or two tabs for example. I've left suggestions for areas that you might want to look at and feel free to ask if you're still confused!!

client/src/pages/About/ExecProfile/ExecProfile.jsx
vproHacks: Changed to 25s!

client/src/pages/About/About.jsx
vproHacks: Added Ternary with constant instead of state

client/src/pages/About/About.jsx
vproHacks: moved to wrapper and changed functionality for specific tab use

client/src/pages/About/About.jsx
vproHacks: Created a default first element by only checking for wanted tabs done without using effect since that would be expensive for each render.

vproHacks:
Added a new variable `wantToLoad` for each tab that controls if that tab is to be loaded. If true then the tab will be loaded, otherwise it will not be rendered.

natapokie (approved):
great job vraj! 
thanks for making these changes, sorry for the unclear instructions in the beginning, hope the changes didn't take too much time!!

for now, I just made a commit to set all the tabs to false :)

PR #617, opened 2023-06-03
603 organize leedur profile dashboard

### LINKED ISSUE #603: Organize Leedur Profile Dashboard
## **Changes:**

Organize all the leedur dashboard permissions in a cleaner format -- it's difficult to navigate with the current format that we set up.
- Here are a few ideas to get started, but you're welcome to think of your own!
  - Use a similar button format and include some [icons](https://searchengineland.com/google-nmx-tips-390006#:~:text=Here%E2%80%99s%20a%20rundown%20of%20the%20changes%3A)
  - Use a similar format to the navbar (vertical or horizontal slider situation with links)

Ideally, try not to change the entire layout of the Profile page, but if you have any ideas please ask @natapokie first! We just want to make sure that the functionality is still there.

This is what the leedur dashboard looks like right now...
Screenshot 2023-05-20 143614

Note: you'll need to run the frontend and backend using docker-compose to complete this task and create a leedur account and request all permissions.

## **Visuals:**

Pretty much up to you! If you need any icons the best place to go is [FontAwesome](https://fontawesome.com/). Just make sure to use svg files so it's easier to change colours as needed.

## **Requirements:**

- [ ] Cleaner and easier-to-navigate UI!
- [ ] Make sure all the links are still there -- i.e., the functionality is the same

### **Documentation:**

- You'll be making changes to the following component, `ProfilePageLeaderPermissionDashboardLinks` in this file, `client\src\pages\Profile\Profile.jsx`
  - you may want to make changes to this class, `profile-leader-dashboard-links`

Tanushanth:
## **Contribution:**
Organized Leedur Profile Dashboard into dropdown menus split into three categories: data, outreach and scunt.

### **Issue:**

Closes #603  

### **Accomplishments:**

- Created DashboardDropdown component (in components folder) to handle an object containing information about each button (label, path, authScopes), which used to be information passed to ProfilePageDashboardLink Component. This was based on the ProfileDropdown Component (Used as reference).
- Created three buttons to act as the dropdown using the dropdown menu mentioned above.

### **Visuals:** 
image
image
image

### **Expected Behavior:**

- The dropdown menus should shrink along with the buttons to be responsive
- They should not overlap (They were, but were fixed), and should be in front of everything else on the screen
- When a button is clicked, it should navigate to the proper URL

### **Testing Steps:**

- Create a Leedur account (using Natalie's guide) with some/all permissions.
- Navigate to a profile page.
- Dropdowns should appear, and be populated with only options that the account has permissions for. (Should test with none/some options)

### **Complications:**

- Started by just organizing buttons, but was suggested by Farbod & Natalie that a Dropdown would work better
- Dropdowns kept overlapping, found out it was because there wasn't enough space in the flexbox, so they would overlap/wrap weirdly (Solution in resources)
- Docker would sometimes freeze and not let me run the backend (mongo-express container specifically), A fix that always worked but took a while was deleting the mongo-volumes & node_modules folders and redoing the setup, but I would need to recreate the Leedur account again.
- Still have difficulty getting the width of the dropdown to always match the button (chose to just make them a size that guarantees it never overlaps)
  - new fix ([7047015](https://github.com/UofT-Frosh-Orientation/orientation-website/pull/617/commits/70470151f1f82030463e383895332cf478659f7a)) saves the width of the main dropdown item (on initial render & window resize), and applies it to the dropdown links/image -- with help from [this link](https://www.pluralsight.com/guides/re-render-react-component-on-window-resize) and [chatgpt](https://chat.openai.com/share/108a8239-f69a-4d3a-8574-33834dccf65a) 😎 

### **Resources:**

Any additional resources used outside this issue.

- Aligning items left/right in a flex div: https://stackoverflow.com/questions/35269947/how-can-i-align-one-item-right-with-flexbox
- Code for CSS grid Tool (Before I changed it to dropdowns): https://cssgrid-generator.netlify.app/
- Overlapping items in flexbox: https://stackoverflow.com/questions/26476697/flexbox-columns-overlap
- Natalie's account creation guide for testing: https://docs.google.com/document/d/1F8wZg6mtSoZ7xe3krBfYsQuPUX1EZYnY5AZmDprPOh8/edit#heading=h.yf8e4qnfhucf
- ProfileDropdown Component Code (To create a dropdown)


client/src/pages/Profile/Profile.jsx
```diff
@@ -509,49 +538,136 @@ const ProfilePageLeaderPermissionDashboardLinks = () => {
...
+        {leader && approved ? (
+          <Link
+            to={'/permission-request'}
+            style={{ textDecoration: 'none' }}
+            className={'no-link-style'}
+          >
+            <Button label="Request Leedur Permissions" />
+          </Link>
+        ) : (
+          <></>
```
natapokie, line 580: you'll need to put all your dropdown divs within this tag so leedurs that are not approved (i.e., `user.approved === false`) u don't see the tags yet,
image

natapokie: you'll need to put all your dropdown divs within this tag so leedurs that are not approved (i.e., `user.approved === false`) u don't see the tags yet,
image

client/src/pages/Profile/Profile.jsx
```diff
@@ -509,49 +538,136 @@ const ProfilePageLeaderPermissionDashboardLinks = () => {
...
+          </div>
+          <DashboardDropdown
+            open={openOutreachDropdown}
+            setOpen={setOpenOutreachDropdown}
+            items={dashboardItems.outreach}
+            title={'outreach'}
+          />
+        </div>
+
+        <div className={'profile-leader-dashboard-scunt-links'}>
```
natapokie: since all three dropdown menus use almost the same code, you can use map and iterate through a list of objects! it would also be best to put the drop downs in a separate folder under the components folder (just to reduce the amount of code in Profile.jsx)

here's one format for your object...similar to the sponsors page, `client\src\util\sponsors.jsx`. 
```jsx
{
  label: "data", // data as an example, it would probably be data, outreach & scunt, this would be used as your key when mapping and also when you use your classnames
  title: "Data", // this can be the title of the dropdown menu
  state: // add the state here
  setState: // add state handler here (or the function that you use in your onClick) 
  icon: // import svg instead of using the tag
}
```

note that since you're using event handlers, you'll have to declare the list of objects within the actual component similar to what you've done for `dashboardItems`.

for importing svgs, you can use them similar to how you use images, using the `` tag, and using `src`, using the `<svg>` tag tends to look a little messy.

you can use this as a reference as well for all the auth scopes, `client\src\pages\ScopeRequest\functions.jsx`

client/src/pages/Profile/Profile.jsx
```diff
@@ -475,30 +480,54 @@ ProfilePageScuntToken.propTypes = {
 
 const ProfilePageLeaderPermissionDashboardLinks = () => {
   const { user } = useSelector(userSelector);
+  const { darkMode, setDarkModeStatus } = useContext(DarkModeContext);
+  const [openDataDropdown, setOpenDataDropdown] = useState(true);
```
natapokie: set the initial state of the dropdowns to false, the dropdowns seem to be initially open
image

client/src/components/DashboardDropdown/DashboardDropdown.scss
```diff
@@ -0,0 +1,78 @@
...
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
+
+  background-color: var(--navbar-bg);
+  color: var(--black);
+  width: 100%;
+  height: 60px;
```
natapokie: instead of using `height`, you can use padding instead to make space around the container! padding just makes sure that things fit better on the screen, and it the container will automatically "grow" depending on the contents inside
image
^ this was using `padding: 8px`

client/src/components/DashboardDropdown/DashboardDropdown.scss
```diff
@@ -0,0 +1,78 @@
...
+  pointer-events: none;
+  opacity: 0;
+}
+
+.dashboard-dropdown-container {
+  position: absolute;
+  display: flex;
+  flex-direction: column;
+  border-radius: 6px;
+  width: max(130px, 13%);
```
natapokie: just a note on the width... i'm trying to figure out a way where you can set the width of the dropdown container according to the width of the "dropdown part of it" since this is fixed, while the other higher dropdown part is like a flex box...

ideally it would be best if the dropdown image and the actual dropdown clicking container were in the same "div", so that way the widths would be the same... but overall it's ok for now, there's quite a bit of refractoring that's needed so we can come back to this later!

(hope that kinda makes sense btw, too many instances of dropdown ahha)

client/src/components/DashboardDropdown/DashboardDropdown.jsx
```diff
@@ -0,0 +1,94 @@
...
+                if (user && user?.authScopes?.approved?.includes(authScope)) {
+                  hasAuthScope = true;
+                  break;
+                }
+              }
+            }
+
+            console.log('STATUS', hasAuthScope);
+
+            const hasAnyRegisterScope =
```
natapokie: instead of hasAnyRegisterScope, it would be best if we don't show the dropdown menus if the user has no scopes in that field (e.g., if user has no scope in scunt, don't show that dropdown), once again this does need some refactoring since it would be a lot easier if the dropdown click part and the dropdown image are in the same component.

we can come back to this later for enhancement unless you're not currently working on another task ;)

client/src/pages/Profile/Profile.scss
```diff
@@ -421,17 +421,125 @@
...
+        .profile-leader-dashboard-data-icon {
+          width: 30px;
+          height: 30px;
+          fill: var(--yellow);
+          margin: auto 0 auto auto;
+          padding-bottom: 2px;
+          padding-right: 5px;
+        }
+        .profile-leader-dashboard-data-text {
+          color: white;
```
natapokie: instead of white, use `var(--text-dynamic)` so that it's still legible in light mode,
image

the yellow icon colour is also a little illegible, but it should be fine for now haha

natapokie (changes requested):
great work tanu! i honestly have no idea how you made the CSS work, the dropdowns fit in so perfectly 👌!

i left a few comments and a few other comments that do involve a little more refactoring with the components -- those ones are larger tasks and might not be the best use of your time right now, so we'll probably create an issue for it and we can get back to it later :)

Freeassassin (approved):
Looks great ty Tanu & Nat!

PR #632, opened 2023-06-12
adding period as special character

### LINKED ISSUE #546: Add period to special character in password
The Regex password validator doesn't count the period "." as a special character. This is a mistake and should be changed.

You will need to make modifications to the following function, `validateFunction`, at the following path `client\src\pages\SignUp\functions.jsx`

natapokie:
## **Contribution:**

Adding period as a special character. This was purely a frontend issue that validates text input.

### **Issue:**

Closes #546  

### **Accomplishments:**

- added period as special character

### **Visuals:** 

None

### **Expected Behavior:**

- In previous versions, you would be unable to create an account (for example), if you used a period as a special character. An error message would show up since it didn't recognize the period
- Now, if you create an account, there would be no error message if you use a period as a special character in your password

### **Testing Steps:**

- Only need to run frontend, `yarn dev`
- Create Account, and try to enter a password with a period "." as the special character, the button should be enabled and not disabled

### **Complications:**

- finding out how to use the match function, it seemed that we just need to add the period character into the square brackets []

### **Resources:**

- [chatgpt](https://chat.openai.com/share/02d6ac1c-5ed0-4fd1-8ec0-24c7fb7f2cde) 😆


client/src/pages/SignUp/functions.jsx
```diff
@@ -11,7 +11,7 @@ export const validateEmail = (email) => {
 
 export function validatePassword(password) {
   return String(password).match(
-    '(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[@$!%*#?&])[A-Za-z0-9@$!%*#?&]{8,}',
+    '(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[.@$!%*#?&])[A-Za-z0-9@$!%*#?&.]{8,}',
```
Sherryzh0830: Should we also account for other special characters here then as well? Such as ~ ` ^ + = ; and so on. I'm just pulling from this [website](https://www.computerhope.com/jargon/s/specchar.htm#:~:text=A%20special%20character%20is%20a,are%20examples%20of%20special%20characters.) for reference regarding what is a special character?

natapokie: > Should we also account for other special characters here then as well? Such as ~ ` ^ + = ; and so on. I'm just pulling from this [website](https://www.computerhope.com/jargon/s/specchar.htm#:~:text=A%20special%20character%20is%20a,are%20examples%20of%20special%20characters.) for reference regarding what is a special character?

great point to mention! you can make a commit to add more of those characters :)

Sherryzh0830 (approved):

Freeassassin (approved):

Issue #333, opened 2022-07-16
Birthday blank when Edit page opened


jameskokoska:

jameskokoska:
It just looks empty, it validates fine.

Issue #330, opened 2022-07-16
Other commute option stays open


jameskokoska:
![image](https://user-images.githubusercontent.com/50821962/179370759-01fd52cd-8af1-4f77-8d55-2e49c7d271ae.png)

If you select yes, then other, then back to no, the Other Commute option stays open

PR #643, opened 2023-06-17
612 registration formik

### LINKED ISSUE #612: Registration Form Update
## **Changes:**

Refactor the Registration form using Formik.
 
## **Visuals:**

No changes required

## **Requirements:**

- [ ] Reformat with Formik
- [ ] Functionality must remain the same

## **External Factors:**

Describe any external requirements that might effect the issue (e.g need to await information from other party).

## **Related Issues/PRs:**

Link any other issue or PR that might be related to this issue.
- #333
- #330
- #619

## **Resources:**

### **Tutorials:**

- [Link 1]()
- [Link 2]()
- [Link 3]()

### **Documentation:**

- [Link 1]()
- [Link 2]()
- [Link 3]()

### LINKED ISSUE #619: Button Component - Button Tag
## **Changes:**

In the button components, change all the `<div>` tag to a `<button>` tag. 

Generally, we should not be using `div` for clickable elements, more information [here](https://www.builder.io/blog/buttons#the-problem-with-divs). In summary, it's mainly due to accessibility purposes, for example when submitting forms usually we can just press the Enter key or Space bar, but we can't do that with a `div`.

Buttons do have "inherent" style, so this is a more CSS-related task, more tips to reset button styling [here](https://www.builder.io/blog/buttons#:~:text=Fixing%20button%20styling)!

You'll need to take a look at the following files...
- client\src\components\button
  - Button
  - ButtonBubble
  - ButtonOutlined

## **Visuals:**

No visual changes are required, just make sure that all buttons have the same functionality and style.

## **Requirements:**

- [ ] Use `<button>` tag instead of `<div>`
- [ ] Make sure all style is the same as before
- [ ] Make sure all functionality is the same as before

## **External Factors:**

None

## **Related Issues/PRs:**

None

## **Resources:**

### **Tutorials:**

- [Button Video](https://youtu.be/lVk6EXvQDkM)
- [Button Article](https://www.builder.io/blog/buttons)

### **Documentation:**

- None

natapokie:
## **Contribution:**

Refactoring registration form
- made changes to checkbox, radio and text input so they use the proper "tag form" (i.e., onChange)
  - no more warnings (yay!) 
- wrapped all components within form -- making them tabbable (improving accessibility)

Made changes to the button component
- also allowing tabbing and focus (improving accessibility) 

Made changes to dropdown component
- added keyboard focus and `onKeyDown` events (so you can tab through the options and press enter to select when focused)

Made changes to button component
- now keyboard focusable (tabs and enter)

Fixed some other small bugs with registration disable fields
- found a small bug that will require a lot of refactoring (not that important though?)
  - if you click Yes commuting, and then Subway (for example), the text box is enabled, but later if you click No not commuting, and click back to Yes, the text box is not enabled
  - this is because we only disable onChange, and since the dropdown with Subway isn't "changing" the textbox is not enabled

### **Issue:**

Closes:
- #612 
- #333 (hopefully fixed? added some local storage stuff)
- #330 (fixed some disabling field logic)
- #619

### **Accomplishments:**

- technically you can fill out the entire form without clicking anything (radio buttons are HTML tags and use arrow keys to change, checkboxes use enter to select)

### **Visuals:** 

No visuals changed

### **Expected Behavior:**

- behaves exactly the same in terms of form submission (all fields in the frosh user are still there)
- clicking Tab will allow you to move through all the form fields
- clicking enter on the last tab for "Pay Now" will take you to the payment page (clicking also works as well)
- button is focusable (if you go to inspect tab you will be able to see keyboard-focusable with a green checkmark)
- bug in #330 is fixed

### **Testing Steps:**

1. create a new account, and fill in all the fields
2. click the button to head over to the registration page and try to fill in all the fields just by using the keyboard
  - since some of the components are HTML tags which have their own functionality (some stuff might be confusing but try it out)
3. test out the [following](https://github.com/UofT-Frosh-Orientation/orientation-website/issues/330) on the Extra Events page
4. submit registration form on last page, and check db to make sure user was properly created

### **Complications:**

- `:focus` is the style when you're focused on that element -- it has inherent style so set `outline: none` in css to customize
- warning where it says something about not unique key -- try using `React.Fragment` tag and put the key into that
- more complications encountered, and solved with links in resources

### **Resources:**

- `<React.Fragment>`: Using <React.Fragment>, you can create a container that doesn't render any additional elements when rendered. It acts as an invisible wrapper and doesn't create any extra nodes in the DOM hierarchy. `key` prop can be applied to the `<React.Fragment>` element surrounding the label and input elements. This ensures that each fragment has a unique key within the loop.
- [Radiobuttons](https://stackoverflow.com/questions/68419302/radio-button-in-react-js-using-map): you need to give each input a unique name attribute, as that's what groups them together. make sure the name you give doesn't repeat, otherwise, you'll have multiple radio groups controlling the same value.
- Making components [keyboard-focusable](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard)
- using [onKeyDown](https://www.tutorialspoint.com/detecting-arrow-key-presses-in-javascript)


client/src/pages/Registration/RegistrationForm.jsx
```diff
@@ -345,64 +361,67 @@ const PageRegistrationForm = ({ editFieldsPage, initialValues, onEditSubmit }) =
...
+                            using your submitted information in order to help plan and deliver
+                            Orientation events
+                          </p>
+                        </b>
+                        <Button
+                          type="submit"
+                          style={{ margin: '0 auto', marginTop: '15px' }}
+                          label={'Pay Now'}
+                          // onClick={handleRegister}
+                          isDisabled={!canRegister}
```
Freeassassin:
```suggestion
                          label={'Pay Now'}
                          isDisabled={!canRegister}
```

Freeassassin (changes requested):
Looks great! Just one thing is that the tabbing through the options has a very slight visual effect. Sometimes I got confused as to what element I was selecting. Maybe make the shadow a bit brighter / larger? Also when you first open the registration page the first field is not automatically selected. This means that the user will have to click on it to start the process.

Also some parts of the Extra Events tab is buggy and I would have to tab multiple times to get to the next element and sometimes the disabled elements would be highlighted. For example:
image
image

idk if you can tell from the picture but the when are you coming to Toronto element is selected even though its disabled.
Also I think that dropdown menus tab through all options even when they are closed.

- [ ] Make selected elements more visible
- [x] Make sure the first element of each page is selected by default
- [ ] Make sure disabled elements cant be selected

Freeassassin (approved):
Looks great! Thank you Nat!!

PR #647, opened 2023-06-25
626 optimize profile page using dynamic imports

### LINKED ISSUE #626: Optimize Profile Page Using Dynamic Imports
## **Changes:**

Use dynamic imports and lazy loading to only load what is necessary in the profile page. Create separate files for all the components inside `profile.jsx`. Group them together to minimize files Based on Frosh and Leedur code.

For example, lines 98:145 of `Profile.jsx` layout the structure of the profile page by relying on a `leader` variable, this should be separated into two components, a `PageProfileFrosh` component or `PageProfileLeadur` component and only one is loaded based on one check of the leader variable.

**NOTE:** The rest of the code frequently checks for the leader variable for logic reasons. These components should be rewritten when separating.
 
## **Visuals:**

Any visuals that are related to these changes

## **Requirements:**

- [x] Watch the video in the **Tutorials** 
- [x] Just copy it :) 
  - [x] Create a `PageProfileFrosh.jsx` and a `PageProfileLeedur.jsx` 
  - [x] Add all the components required to both and make sure that there is no checks based on if the userType is leadur or not
  - [x] Dynamically add both to `Profile.jsx` and check the leadure variable onces to determine which one needs to be loaded
- [ ] Time how long it takes to load the profile page with the browser dev tools and see if its actually improving or not (mostly for fun lol I dont care if the times are not different. this whole issue is more for production anyway)

## **Related Issues/PRs:**

Link any other issue or PR that might be related to this issue.

## **Resources:**

### **Tutorials:**

- [Dynamic Import Tutorial](https://youtu.be/JU6sl_yyZqs)

### **Documentation:**

- [Article on Dynamic Imports](https://blog.logrocket.com/react-dynamic-imports-route-centric-code-splitting-guide/)

uzFer:
### **Issue:**
Closes #626 

### **Accomplishments:**
- Split the code into two files: `PageProfileFrosh.jsx` and `PageProfileLeedur.jsx` where `Profile.jsx` checks if the user is a leader or frosh and loads the correct page accordingly.
- In each component, there were many if (leedur === true) checks. I was able to get rid of most of those since the pages are now separately loaded.
- There was a `ProfileHeader` component that would take in the boolean props: `leedur` and `editButton`. I put this component in both `PageProfileFrosh.jsx` and `PageProfileLeedur.jsx` files and removed the leedur prop from both since its obvious which header is for which user now. 
- Since the file names changed, I had to update some imports from two files: `ScuntHome.jsx` and `ProfileEdit.jsx`

### **Visuals:** 
Same old profile for Frosh (I only loaded frontend hence missing info): 
image

Same old profile for Leader (I only loaded frontend hence missing info): 
image

### **Expected Behavior:**
- You should be able to just log in or sign up as Frosh/Leader and access all components as necessary.
- All components should work and show up on your profile page including header, edit button, schedule, scunt team/token, etc.

### **Testing Steps:**

With backend:
- Run docker
- Head to sign up or log in page (uncomment login page in `util/pages.jsx` if need be) with your Frosh account
- Access the Profile page and see if any component is missing
- Do the same steps with a Leader account as well

Frontend only:
- Go to `util/pages.jsx` and uncomment the `<AuthorizedPage></AuthorizedPage>` tags for the `Profile.jsx` page
- For Frosh account: just run the front end
- For Leader account: change to `const leader = true` on line 15
- Access the Profile page and see if any component is missing

### **Complications:**
- Couldn't figure out the dynamic importing since Suspense and lazy were giving me errors like this one
image
- The files are still pretty long with ~700 lines of code. If needed, I can further split the components
- Since I was using just front end to check, I'm not sure if all components are properly shown such as scunt teams/QR codes
- I think the leader dropdown that was recently merged is different from my working copy, so let me know if I need to steal that one and work on editing it

### **Resources:**
- Tried this for debugging lazy loading but no luck yet: [https://stackoverflow.com/questions/58297907/suspense-is-not-loading-content-after-render-reactjs](url)


client/src/components/profile/ProfilePageResources/ProfilePageResources.jsx
natapokie, line 1: if you can add the resources and schedule as a separate component since they're used for both leedur and frosh pages!

client/src/util/pages.jsx
```diff
@@ -60,12 +60,12 @@ export const pages = {
     // },
   ],
   special: [
-    //   {
```
natapokie: uncommented this for now!

server/src/controllers/UserController.js
natapokie, line 1: just want to confirm that the UserController and UserServices is part of the pr or whether it's from the email confirmation pr?

natapokie (changes requested):
great work on separating all the components uzma! 

i merged the branch with dev, so you have the updated version (thought it would be easier since i could run docker ahha)!

as for dynamic loading, you can try using this package that's already installed called.... [react-lazy-load-image-component](https://www.npmjs.com/package/react-lazy-load-image-component), and you can use `LazyLoadComponent` tag, I'm honestly not sure if it actually helps with performance, but we use the same lazy loading for images so i thought it might be good to try!

As for splitting up the components more, I got started on a few components that were integrated with the backend... You can use the folder under components called profile, made a few as a template that you can follow!

For components like resources and schedule, since used in both frosh and leedur profiles (they should be frontend stuff only too), it would be great if you can put them in separate components!

(Also I think i left a ton of commented-out code, so if you notice anything please help me delete, thanks!!)

server/src/controllers/UserController.js
uzFer: should be just from the email confirmation pr, I see those files were changed here too - don't remember why I did but probably a spacing or console.log thing. Also changed the sponsors.md documentation file because I noticed a typo I made from before

natapokie: ok sounds good! just wanted to check, will resolve this convo

natapokie (approved):

Tanushanth (approved):
All good! I tried with leader and normal accounts and everything comes up file on the profile page, only thing, I had to add this at the top of LeedurServices (but it exists in the main branch so not sure if we need to worry abt it) 

const emailConfirmationSubscription = require('../subscribers/emailConfirmationSubscription');

Freeassassin (approved):
Great work!

PR #690, opened 2023-07-22
679 fix frosh assignment

### LINKED ISSUE #679: Fix frosh assignment
## **Expected Behavior:**
Too many alphas
- writing a script and mapping it to a button in the frontend

## **Requirements:**

- [x] Add function to `FroshServices.js` that can retrieves all frosh OR a set of frosh based on search query (see `getFilteredFroshInfo()`) and redistributes them among all frosh groups
- [x] Create routes for both of the above cases. Make sure that it requires `"admin:all"` privilege to access. 
- [x] Add an admin page in frontend for frosh redistribution. 
  - Should just be one button for now.
- [x] Hook up the button to the backend using redux

Tanushanth:
### **Issue:**

Closes #679 

### **Accomplishments:**

- Created frosh route in `froshRoutes.js` for /reassign that uses the `hasAuthScopes` middleware to ensure that the 'admin:all' auth scope is present before allowing the user to reassign users.
- Created a function in FroshController called `reassignFrosh` that grabs all registered frosh(copy of getFilteredFroshInfo), and then recursively loops through and assigns new frosh groups to each user fetched if they have capital pronouns (ex She/Her instead of she/her). The recursion makes sure the same user will not get placed in the same group again.
- Created a simple page at `/frosh-redistribution` with a button to reassign users hooked up to redux (Made redistrubuteFroshSaga in /state/frosh/saga)

### **Expected Behavior:**

- When the button is clicked (after it is made) it should shuffle up the frosh registered users into different groups based on the getNewFroshGroup function

### **Testing Steps:**

- Create a leedur account, and manually add the `admin:all` scope so that you can access the `/frosh-redistribution` page
- Create some registered frosh to test the redistribution on, and manually set their pronouns to capitals (she/her -> She/Her)
- Navigate to /frosh-redistribution and click the button, and see if the frosh have been distributed to different groups.

### **Complications:**

- Still not sure if recursion is the best way to loop through the frosh and make sure they get new groups
- Still need to test with multiple registered frosh to ensure that it is shuffling correctly

### **Resources:**

- resource 1
- resource 2
- resource 3

### DIFF
--- a/server/src/controllers/FroshController.js
+++ b/server/src/controllers/FroshController.js
@@ -134,6 +134,26 @@ const FroshController = {
 
   async reassignFrosh(req, res, next) {
     try {
+      async function mapFroshUsers(frosh, index) {
+        if(index >= frosh.length){
+          return;
+        }
+        const froshUser = frosh[index];
+        const { froshGroup, froshGroupIcon } = await FroshServices.getNewFroshGroup(
+          froshUser.discipline,
+          froshUser.pronouns,
+        );
+        if(froshGroup == froshUser.froshGroup){
+          //Redo this index
+          await mapFroshUsers(frosh, index);
+        }
+        else{
+          //Change frosh group of this user
+          await FroshServices.updateFroshInfo(froshUser.id, {froshGroup: froshGroup, froshGroupIcon: froshGroupIcon});
+          await mapFroshUsers(frosh, index + 1);
+        }
+      }
+      
       if (!req.user?.froshDataFields?.approved?.length) {
         return next(new Error('UNAUTHORIZED'));
       }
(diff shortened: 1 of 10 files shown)


server/src/controllers/FroshController.js
```diff
@@ -130,6 +130,40 @@ const FroshController = {
...
+        return next(new Error('UNAUTHORIZED'));
+      }
+      const filter = req.user?.froshDataFields?.approved.reduce(
+        (prev, curr) => {
+          prev[curr] = 1;
+          return prev;
+        },
+        { _id: 1 },
+      );
+      query = {isRegistered: true };
```
Freeassassin:
```suggestion
      const query = {isRegistered: true };
```

Freeassassin (approved):
Looks great, thank you for the help with this. 

I added some quick changes to fix the issues.

Issue #624, opened 2023-06-05
[draft] Stop Registration!


natapokie:
## **Changes:**

After registration closes, hide the "Create Account" button and replace it with "Registration Closed".

There is already some logic in the code that does this, in the following file... [client\src\pages\SignUp\SignUp.jsx](https://github.com/UofT-Frosh-Orientation/orientation-website/blob/dev/client/src/pages/SignUp/SignUp.jsx) starting on lines 203. 

We want to avoid commenting out the code in the future, so similar to the Landing page ["wrapper"](https://github.com/UofT-Frosh-Orientation/orientation-website/pull/610#:~:text=added%20wrapper%20with%20variable%20readyForFrosh%20to%20control%20what%20is%20displayed)

## **Visuals:**

Any visuals that are related to these changes

## **Requirements:**

- [ ] Display the registration closed message depending on a variable (e.g. closeRegistration) that is set to true/false

## **External Factors:**

Describe any external requirements that might effect the issue (e.g need to await information from other party).

## **Related Issues/PRs:**

Link any other issue or PR that might be related to this issue.

## **Resources:**

Link any resources or people that might be helpful in the accomplishment of the issue.

rohit-yadavv:
i want to work on this please assign it to me

RohinMarok:
Hello, I want to fix the issue so please assign it to me so I can work on it.

AbhayWalia:
Can I work on this issue please. I will try have it done ASAP.

RohinMarok:
I can work on this issue will send a pr request soon

AbhayWalia:
No Rohin this one is mine

PR #610, opened 2023-05-29
608 load up landing page

natapokie:
## **Contribution:**

- Looping through the created landing pages (currently Tanu, Uzma and mine from last year), to render a different landing page on every refresh using local storage keys.
- added wrapper with variable `readyForFrosh` to control what is displayed

### **Issue:**

Closes #608 

### **Accomplishments:**

- set up local storage key `landing_page_idx` (will show up in dev tools F12 tab under Application > Local Storage > Key)
- generates a "random" local storage key each time, which indexes a certain landing page in the defined array

### **Visuals:** 

- Shows all landing pages from previous issues #598 

### **Expected Behavior:**

- upon first loading, you'll see one landing page, and after refreshing you'll see a different landing page each time
- landing pages should look exactly the same as the ones from the prs

### **Testing Steps:**

Steps required to get expected behavior.

- run the frontend (no need to use docker compose)
- landing page will show up as it's rendered in App
- refresh to check for different landing pages

### **Complications:**

- needed to learn how to use local storage again
- needed to alter useEffect to prevent rerenders when the local storage key was updated
- called components incorrectly and got an error `Functions are not valid as a React child`
- arrays need keys or eslint throws error
- figuring out how to generate random numbers 
- react is a little funky with indexing arrays -- had to use `map()`, if anyone else has ideas let me know

### **Resources:**

Any additional resources used outside this in issue.

- [resource 1](https://javascript.info/localstorage): docs on local storage
- [resource 2](https://github.com/UofT-Frosh-Orientation/orientation-website/blob/dev/client/src/pages/ScuntJudges/ScuntJudges.jsx): code example from code base on local storage
- [resource 3](https://stackoverflow.com/questions/4959975/generate-random-number-between-two-numbers-in-javascript): generate a random number between two given numbers
- [resource 4](https://bobbyhadz.com/blog/react-functions-are-not-valid-as-react-child): take a look for this error `Functions are not valid as a React child`

PR #755, opened 2023-12-14
Addressed Issue Stop Registration! 

RohinMarok:
## **Contribution:**

I added the stop registration button which doesn't allow users to register after it has been closed. I added the button to work dynamically based on the data returned from an API call effectively conditionally rendering the button based on the team's requirements (this part is commented out in the file but you can comment it back in and route your API calls).

### **Issue:**

After registration closes, hide the "Create Account" button and replace it with "Registration Closed".

There is already some logic in the code that does this, in the following file... [client\src\pages\SignUp\SignUp.jsx](https://github.com/UofT-Frosh-Orientation/orientation-website/blob/dev/client/src/pages/SignUp/SignUp.jsx) starting on lines 203.

We want to avoid commenting out the code in the future, so similar to the Landing page https://github.com/UofT-Frosh-Orientation/orientation-website/pull/610#:~:text=added%20wrapper%20with%20variable%20readyForFrosh%20to%20control%20what%20is%20displayed

### **Accomplishments:**

- edited the home page to implement this functionality

Screenshot 2023-12-13 at 7 02 16 PM


RohinMarok:
Fixes #624

jameskokoska:
ECE444?

PR #784, opened 2024-06-24
removed WhatsApp link

kexinxiang:
## **Contribution:**

Commented out the WhatsApp link in util/socials.jsx since we are not using WhatsApp for 2T4 Frosh


gaurikam2003 (approved):
great work and thanks!

PR #812, opened 2024-07-16
Fix cursor on schedule buttons

kexinxiang:
## **Contribution:**

Cursor stay in style when hovering on the schedule buttons

### **Visuals:** 

Any related visuals.

IMG_4804


ashleyleal (approved):
Perfect! Thank you :D

PR #859, opened 2025-06-27
added medals to sponsor cards

pengkev:
for some reason only on the diamond card when you hover you'll see purple but not diamond color.


nrakhshan:
Lol that's because the colour is set like that in the CSS, I'll change it to blue

nrakhshan (approved):

nrakhshan:
Oh strange you did change it hmmm

nrakhshan:
ok figured it out

PR #871, opened 2025-07-04
Center FAQ buttons when screen is large

alyssaHee:
Centered FAQ buttons once the screen is large enough to not need the scrollbar


nrakhshan (commented):
omg tysm you've extended my lifespan by 10 years

nrakhshan (approved):

PR #873, opened 2025-07-07
Edit account for unregistered profile fix

alyssaHee:
Made the already filled out information show in the edit account page, hopefully no other text input got altered through my changes.


nrakhshan:
I finally stopped procrastinating on this ksdkfdsf but idk why it breaks the api i think this will have to be an issue for next years tech team to resolve <//3

PR #884, opened 2025-08-18
This change most deals with the bland login page to make it look more exciting

pengkev:
stole the idea of overlaying assets ontop to make a 3d version, also changed the ordering of the html tags for the login box to be centered after rendering background

commit to put qr on bottom for mobile but this is unverified


nrakhshan:
Kevin I'm sorry, this is great but I don't know if this fits with the rest of the website 😭😭😭 it's a little out of nowhere it's not gonna look coherent with the rest of the visuals and colours <//3

pengkev:
understandable i had trouble getting images that would fit that was the best one

